PR #18, opened 2019-02-15
Add support "unsafe to use" endpoints

macisamuele:
This PR exposes one of the internally developed features to ensure that endpoints that are _not good enough_ to be "blindly" used from users of the generated code.

All the endpoints that are annotated with ``x-unsafe-operation`` will be annotated with a deprecation warning.

The unsafe operations could be defined with two different approaches:
 * in ``#/info/x-operation-ids-unsafe-to-use`` as a list of operation ids
 * on the operation object itself with ``x-unsafe-operation: true``

@cortinico : Could you please review this and give me info on where to write to write tests and documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,11 @@
 .PHONY: install-hooks
 
+.git/hooks/pre-commit: venv
+	${CURDIR}/venv/bin/pre-commit install --install-hooks
+
+install-hooks: .git/hooks/pre-commit
+	@true
+
 venv:
 	virtualenv venv
 	./venv/bin/pip install pre-commit
-
-install-hooks: venv
-	./venv/bin/pre-commit install --install-hooks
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -1,8 +1,23 @@
...
+install-hooks: ${CURDIR}/.git/hooks/pre-commit
+	@true
+
+build:
+	${GRADLE} plugin:build
 
-install-hooks: venv
-	./venv/bin/pre-commit install --install-hooks
+generateSwagger assembleDebug check:
+	${GRADLE} $@
```
cortinico: Can we clean this up a bit?
This https://github.com/gradle/gradle-completion looks like a great alternative ;)

macisamuele: Definitely correct.
As you noticed the makefile targets were mainly created to have tab completion. 
I'm going to remove the commit

PR #22, opened 2019-03-20
Add supporrt for composed models and reference models

macisamuele:
The goal of this PR is to support specifications that have:
 * reference to references
```yaml
definitions:
  model:
    type: object
    properties: ...
  model_r:
    $ref: '#/definitions/model'
 ```
 * definitions that have `allOf` attributes where only one item does characterise the final schema
```yaml
definitions:
  model:
    type: object
    allOf:
    - $ref: '#/definitions/something'
    - description: "override the description"
 ```


macisamuele:
@cortinico ping

PR #23, opened 2019-03-21
Read spec version from swagger spec via codegen tools

macisamuele:
The idea of this PR is to ensure that version extraction from swagger specs supports YAML format too.

In order to avoid adding a new dependency (`org.yaml.snakeyaml`) and manually reading the file I've opted to delegate the parsing process to `SwaggerParser` provided by swagger-codegen.

NOTE: The check in `readVersionFromSpecfile` is not needed in case the specs are valid as `#/info/version` is required. As we don't revalidate the specs adding the check gives us the flexibility of supporting slightly invalid specs as well

### DIFF
--- a/plugin/src/main/java/com/yelp/codegen/plugin/CodegenPlugin.kt
+++ b/plugin/src/main/java/com/yelp/codegen/plugin/CodegenPlugin.kt
@@ -1,6 +1,7 @@
 package com.yelp.codegen.plugin
 
 import com.yelp.codegen.main
+import io.swagger.parser.SwaggerParser
 import org.gradle.api.DefaultTask
 import org.gradle.api.plugins.BasePlugin
 import org.gradle.api.tasks.Input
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/com/yelp/codegen/plugin/CodegenPlugin.kt
```diff
@@ -126,15 +123,16 @@ open class GenerateTask : DefaultTask() {
...
+        val swaggerSpec = SwaggerParser().readWithInfo(specFile.absolutePath, listOf(), false).swagger
+        specVersion = when (val version = swaggerSpec.info.version) {
+            is String -> {
                 println("Successfully read version from Swagger Spec file: $version")
-                specVersion = version
+                version
+            }
+            else -> {
+                println("Issue in reading version from Swagger Spec file. Falling back to 0.0.0")
+                "0.0.0"
```
cortinico:
```suggestion
                "DEFAULT_VERSION"
```

PR #26, opened 2019-05-08
Fix multi part issue

macisamuele:
The goal of this PR is to address an issue related to file uploads.
The issue was identified internally and ticketed on COREMOBAPI-1461

The issue could be summarised as: _we're generating incorrect code in case of file parameters_

Assuming a simple endpoint
```json
"/send/file": {
    "post": {
        "consumes": ["multipart/form-data"],
        "parameters": [
            {
                "in": "formData",
                "type": "file",
                "name": "file",
                "required": true
            }
        ],
        "responses": {
            "200": {
                "description": "fafa"
            }
        }
    }
}
```

The generated code should look like
```kotlin
@JvmSuppressWildcards
interface DefaultApi {
  @retrofit2.http.Multipart    // <- different
  @Headers(
    "X-Operation-ID: "
  )
  @POST("/send/file")
  fun sendFilePost(
    @retrofit2.http.Part(    // <- different
      "file\"; filename=\"file") file:  RequestBody): Completable
}
```
while the following code is the currently generated content
```kotlin
@JvmSuppressWildcards
interface DefaultApi {
  @retrofit2.http.FormUrlEncoded
  @Headers(
    "X-Operation-ID: "
  )
  @POST("/send/file")
  fun sendFilePost(@retrofit2.http.Field("file\"; filename=\"file") file:  RequestBody): Completable
}
```

The issue is caused by the fact that swagger-codegen does not keep in consideration the operation level `consumes` key to properly set `isMultipart` attribute of `Operation` object.

### DIFF
--- a/plugin/src/main/java/com/yelp/codegen/SharedCodegen.kt
+++ b/plugin/src/main/java/com/yelp/codegen/SharedCodegen.kt
@@ -374,9 +374,7 @@ abstract class SharedCodegen : DefaultCodegen(), CodegenConfig {
      * Determine if the swagger operation consumes mutipart content.
      */
     private fun isMultipartOperation(operation: Operation?): Boolean {
-        return operation?.consumes != null && operation.consumes.any {
-            consume -> consume == "multipart/form-data"
-        }
+        return operation?.consumes?.any { it == "multipart/form-data" } ?: false
     }
 
     /**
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/com/yelp/codegen/SharedCodegen.kt
```diff
@@ -370,6 +370,15 @@ abstract class SharedCodegen : DefaultCodegen(), CodegenConfig {
         return codegenModel.dataType
     }
 
+    /**
+     * Determine if the swagger operation consumes mutipart content.
+     */
+    private fun isMultipartOperation(operation: Operation?): Boolean {
+        return operation?.consumes != null && operation.consumes.any {
```
cortinico:
```suggestion
        return operation?.consumes?.any { it == "multipart/form-data" } ?: false
```

PR #28, opened 2019-05-20
Added test to explain usage of Swagger Codegen with Retrofit

### LINKED ISSUE #27: Read Swagger.json from URL
Can Specification be read from a URL ?

Also Do you have any sample code of using these with Rx and Retrofit?

cortinico:
I've added a JUnit test where a Retrofit instance is initialised and used to talk to Petstore instance.

Fixes #27 (ping @yudhir)

### DIFF
--- a/samples/generated-code/simplified_pet_store.json
+++ b/samples/generated-code/simplified_pet_store.json
@@ -69,13 +69,12 @@
         }
     },
     "info": {
-        "basePath": "/v2",
         "description": "This is a simplified version of the sample server Petstore server.",
         "title": "Swagger Petstore",
         "version": "1.0.0"
     },
     "paths": {
-        "pet/{petId}": {
+        "/pet/{petId}": {
             "get": {
                 "description": "Returns a single pet",
                 "operationId": "getPetById",
--- a/samples/generated-code/simplified_pet_store.json
+++ b/samples/generated-code/simplified_pet_store.json
@@ -69,12 +69,13 @@
         }
     },
     "info": {
+        "basePath": "/v2",
         "description": "This is a simplified version of the sample server Petstore server.",
         "title": "Swagger Petstore",
         "version": "1.0.0"
     },
     "paths": {
-        "/pet/{petId}": {
+        "pet/{petId}": {
             "get": {
                 "description": "Returns a single pet",
                 "operationId": "getPetById",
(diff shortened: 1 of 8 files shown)


samples/generated-code/simplified_pet_store.json
```diff
@@ -69,12 +69,13 @@
...
     },
     "info": {
+        "basePath": "/v2",
         "description": "This is a simplified version of the sample server Petstore server.",
         "title": "Swagger Petstore",
         "version": "1.0.0"
     },
     "paths": {
-        "/pet/{petId}": {
+        "pet/{petId}": {
```
macisamuele: Please do not remove the initial `/` from the path

samples/generated-code/src/test/java/com/yelp/codegen/generatecodesamples/InvokeRetrofitTest.kt
```diff
@@ -0,0 +1,26 @@
...
+import retrofit2.adapter.rxjava2.RxJava2CallAdapterFactory
+
+class InvokeRetrofitTest {
+
+    @Test
+    fun createAndInvokeRetrofit() {
+        val retrofit: Retrofit = Retrofit.Builder()
+                .addCallAdapterFactory(RxJava2CallAdapterFactory.create())
+                .addConverterFactory(GeneratedCodeConverters.converterFactory())
+                .baseUrl("https://petstore.swagger.io/v2/")
```
macisamuele: As far as I might like this idea, what is going to happen if `petstore.swagger.io` is offline? I do expect that our tests will start to fail.
Additionally this type of dependency forces developers to have internet connection while running tests which is not necessarily great.

My proposal would be to have a tiny mock server that catches requests and return "pre-defined" answers.
NOTE: I'm not looking for a complex system that is able to filter based on HTTP verb, path, parameters, etc. but rather something that "if I'm called then no mater what the answer is X" 😄

samples/generated-code/src/test/java/com/yelp/codegen/generatecodesamples/InvokeRetrofitTest.kt
```diff
@@ -0,0 +1,26 @@
...
+                .build()
+
+        val defaultApi = retrofit.create(DefaultApi::class.java)
+        val pet = defaultApi.getPetById(1).blockingGet()
+
+        println(pet)
```
macisamuele: Remove the print

PR #29, opened 2019-05-24
Kotlin Coroutines Support

cortinico:
This is an initial PR for supporting another template to generate Kotlin-coroutines code.
I'd love to get some early feedback from @macisamuele before I continue working on this to decide how we want to setup the project.

I'm not a big fan of copying all the `.mustache` files over, so we should probably re-organize them in subfolders?

What's missing:
- [x] Tests
- [x] Updated the readme
- [x] Add a `kotlin-coroutines` to the samples

### DIFF
--- a/plugin/src/main/java/com/yelp/codegen/SharedCodegen.kt
+++ b/plugin/src/main/java/com/yelp/codegen/SharedCodegen.kt
@@ -36,6 +36,7 @@ const val HEADERS_TO_IGNORE = "headers_to_ignore"
 internal const val X_NULLABLE = "x-nullable"
 internal const val X_MODEL = "x-model"
 internal const val X_OPERATION_ID = "x-operation-id"
+internal const val X_FUNCTION_QUALIFIER = "x-function-qualifier"
 internal const val X_UNSAFE_OPERATION = "x-unsafe-operation"
 
 // Headers Names
--- a/plugin/src/main/java/com/yelp/codegen/SharedCodegen.kt
+++ b/plugin/src/main/java/com/yelp/codegen/SharedCodegen.kt
@@ -36,7 +36,7 @@ const val HEADERS_TO_IGNORE = "headers_to_ignore"
 internal const val X_NULLABLE = "x-nullable"
 internal const val X_MODEL = "x-model"
 internal const val X_OPERATION_ID = "x-operation-id"
-internal const val X_FUNCTION_QUALIFIER = "x-function-qualifier"
+internal const val X_FUNCTION_QUALIFIERS = "x-function-qualifiers"
 internal const val X_UNSAFE_OPERATION = "x-unsafe-operation"
 
 // Headers Names
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ Here the list of the supported platforms:
 | Platform | Description                                |
 | -------- | ------------------------------------------ |
 | `kotlin` | Generates Kotlin code and Retrofit interfaces, with RxJava2 for async calls, Moshi for serialization and ThreeTenABP for Data management |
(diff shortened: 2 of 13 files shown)


macisamuele:
@cortinico we already discussed about the PR in person, but it would be better to keep a trace over here as well.

I do like the idea of having a new "template" that allows generation of coroutine based code.
Duplicating code and templates as provided in this PR could be a possibility but I do expect higher maintenance costs.

What if we do have a generic `KotlinGenerator` that will be extended by two independent generators (ie: `KotlinRxJavaGenerator` and `KotlinCoroutineGenerator`)?
By doing so we can ensure that all the kotlin specific details are defined once and reused.

The needed additions would be:
 * having a method that given the output type generates the appropriate string (ie. `Type` -> `Single<Type>|Completable|Defer<Type>|Type` according to the case)
 * templating the method modifiers to allow injection of `suspend`  keyword in the API methods generation

NOTE: Doing something like this will then require a major release as the original currently existing template would not be usable anymore

cortinico:
After some discussion with @macisamuele we feel more comfortable pushing this to a future release

chathudan:
@cortinico are you planning to release this sooner?

cortinico:
> @cortinico are you planning to release this sooner?

I'm currently looking for feedback on this. I would be interested in understanding if you're planning to use this modified template.
Adding coroutines support is quite straightforward, but it's yet another template that we have to maintain. Having active users that are the template is a valid reason to spend effort on it.

Regarding the timeline, ideally we can manage to have this merged and have a new release by the end of November.

StevenKek:
I would love to use this, but I need it ASAP. Currently I'm using Swagger Codegen which has Kotlin coroutine support already.
I can't, however, force it to use ThreeTenABP LocalDate type instead of string for date fields. And it seems it will not be resolved in the foreseeable future :(

Is there a way to generate code with this pull request? I don't know how to download entire repository, patched with a specific PR out of GitHub. Also this codegen is used through Gradle I noticed. That makes it even more difficult for me then. At least a believe so. I'm new to Android Studio, Gradle and all, sorry.

Thanks

cortinico:
Hi all, please find a production-ready version of the coroutines support. Other than @macisamuele 's feedback I'd love to get an ack from either @redwarp or @filipemp.

I've took a different approach from the first commit. We now don't need to clone the entire `.mustache` file hierarchy. 

> @cortinico are you planning to release this sooner?

This will be in the upcoming release @chathudan, unless we find a major blocker during the review.

> I would love to use this, but I need it ASAP. 

@StevenKek apologize for the delay but I took a 2 months sabbatical and I haven't really checked my Github notifications :/  Ideally we're going to ship this soon and you'll have it bundled in the upcoming release.

> Is there a way to generate code with this pull request?

You could checkout the fork and install the gradle plugin inside your maven local. It's doable but a bit complicated to setup. If you're not in a hurry, I advice you wait for the upcoming release.

plugin/src/main/java/com/yelp/codegen/SharedCodegen.kt
```diff
@@ -36,6 +36,7 @@ const val HEADERS_TO_IGNORE = "headers_to_ignore"
 internal const val X_NULLABLE = "x-nullable"
 internal const val X_MODEL = "x-model"
 internal const val X_OPERATION_ID = "x-operation-id"
+internal const val X_FUNCTION_QUALIFIER = "x-function-qualifier"
```
macisamuele: I would use the plural instead of the singular, as you can have multiple qualifiers to add to a method.

cortinico: Yup, fixed

cortinico:
> Very nice approach to the problem.
> Reduce template duplication while preserving consistency and expandability!

Awesome! Thanks for the approval.
Let's wait for an ack from either @redwarp or @filipemp

PR #42, opened 2019-07-10
Add Integration tests for responses 

cortinico:
This commit introduces multiple JUnit tests for sending and receiving
responses using the Swagger Codegen generated models.

Tests are running using the OkHttp MockWebServer and mock responses
have been inlined in the body of the tests.


samples/junit-tests/src/test/java/com/yelp/codegen/generatecodesamples/FormatEndpointTest.kt
```diff
@@ -0,0 +1,92 @@
...
+    @get:Rule
+    val rule = MockServerApiRule()
+
+    @Test
+    fun formatEndpoint_withEnumFormat() {
+        rule.server.enqueue(MockResponse().setBody("""
+            {
+                "enum_property": "VALUE1"
+            }
+        """.trimIndent()))
```
macisamuele: nit. is this needed? I think that this is not needed and eventually I would suggest to remove it as it does not add any value to the test

cortinico:
> The tests do _check_ that the decoding of the response is working as expected but we're not doing much with respect to encoding of the request. Eventually doing it in a follow-up would be great considering the extent of the current PR.

Agree, I'd rather keep it outside of this PR anyway as it's already pretty big.
I can extend every test to make sure that the requests received by the `MockWebServer` are properly encoded.

samples/junit-tests/src/test/java/com/yelp/codegen/generatecodesamples/FormatEndpointTest.kt
cortinico: `is this needed?` -> What are you referring to?
The MockResponse here is necessary as otherwise the test will fail.

The `.trimIndent()` is automatically added by IntelliJ when opening a multiline string and makes sure the strings are well trimmed. I can eventually remove all of them but I don't think they hurt so much.

macisamuele: I wasn't aware about the intellij smart additions

macisamuele:
Feel free to merge as is and eventually follow up with _requests received by the MockWebServer are properly encoded_

PR #48, opened 2019-07-11
Add JUnit Tests for the plugin classes

cortinico:
I'm adding multiple JUnit tests for the generator and Util classes inside the `plugin` module.

e298a23 Make sure Detekt runs on tests as well
c9a36b4 Add JUnit tests for KotlinGenerator
9857ec7 Add JUnit tests for KotlinLangUtils
9d6a86d Add JUnit tests for StringUtils

### DIFF
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -50,7 +50,7 @@ pluginBundle {
 
 detekt {
     toolVersion = "1.0.0-RC16"
-    input = files("src/main")
+    input = files("src/")
     config = files("./detekt-config.yml")
     buildUponDefaultConfig = true
     filters = ".*/resources/.*,.*/build/.*"
--- a/plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
+++ b/plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
@@ -2,6 +2,7 @@
 
 package com.yelp.codegen
 
+import com.google.common.annotations.VisibleForTesting
 import com.yelp.codegen.utils.KotlinLangUtils
 import com.yelp.codegen.utils.safeSuffix
 import com.yelp.codegen.utils.sanitizeKotlinSpecificNames
@@ -62,7 +63,8 @@ class KotlinGenerator : SharedCodegen() {
      * ABSTRACT FIELDS AND CONFIG FUNCTIONS
      ==================================================== */
 
-    override val platform = "android"
+    @VisibleForTesting
+    public override val platform = "android"
 
     override fun getName() = "kotlin"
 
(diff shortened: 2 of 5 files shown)


plugin/build.gradle.kts
```diff
@@ -50,7 +50,7 @@ pluginBundle {
 
 detekt {
     toolVersion = "1.0.0-RC16"
-    input = files("src/main")
+    input = files("src/")
```
macisamuele: ++
Nice catch

plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
```diff
@@ -131,13 +133,16 @@ class KotlinGenerator : SharedCodegen() {
     /** No testing files are needed on Kotlin Generator */
     override val testingSupportFiles = listOf<SupportingFile>()
 
-    override fun listTypeWrapper(listType: String, innerType: String) =
+    @VisibleForTesting
+    public override fun listTypeWrapper(listType: String, innerType: String) =
```
macisamuele: kotlin default visibility is `public`, this change is making them private (according to the [doc](https://developer.android.com/reference/android/support/annotation/VisibleForTesting)).
Is this change an intentional change?
If so, then this change _breaks_ `KotlinGenerator` backward compatibility.

Options here are:
 * do not make the method private if not used in testing (~= revert this diff)
 * have a major version bump
 * have a minor bump highlighting the fact that there is a breaking change. NOTE: this option is actually going a bit beyond what sem-versioning defines but I do assume that the method was never assumed to be public

plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
```diff
@@ -176,7 +182,8 @@ class KotlinGenerator : SharedCodegen() {
         return codegenModel
     }
 
-    private fun addRequiredImports(codegenModel: CodegenModel) {
+    @VisibleForTesting
+    internal fun addRequiredImports(codegenModel: CodegenModel) {
```
macisamuele: According to the [doc](https://developer.android.com/reference/android/support/annotation/VisibleForTesting) `@VisibleForTesting` without further specifications makes the decorated method private.

How is `internal` going to be used here?

plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
cortinico: > kotlin default visibility is public, this change is making them private (according to the doc).

Yup, default visibility is public, but not in this case.
Here the visibility is inherited from the super class (is `protected`) so I'm relaxing it to `public`. 

Given that the visibility is wider I don't think this is a backward incompatible change.

plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
cortinico: > According to the doc @VisibleForTesting without further specifications makes the decorated 
> method private.

Not really.
The doc says that you can optionally specify which would have been the desired visibility. If you don't specify the desired visibility, it will default to private:

> If not specified, the intended visibility is assumed to be private.

Anyway this desired visibility has no real effect on the bytecode and is mostly used by tools (e.g. the IDE) to show warnings on usages of those fields.

PR #50, opened 2019-07-13
Remove leading slash if Swagger Spec are specifying a basePath

### LINKED ISSUE #49: Endpoint URL is invalid
@cortinico This is with regards to #46 , 

My API base URL is something like this  ``` https://example.com/api/v5/```  and Retrofit required "/" at the end of the base URL. 

Since the plugin add "/" to API `@GET("/something")`  so it goes to 
```@GET("/something"): https://example.com/something```

BUT actually, it should go to  ```https://example.com/api/v5/something``` which means the API interface should remove "/"  ```@GET("something")``` . I removed "/" **manually** and the generated SDK working fine. 

How can I resolve this rather modify the generated APIs manually? 

NB: 

> Add trailing slash to base URL, remove leading slash from relative URL.

https://github.com/square/retrofit/issues/907#issuecomment-141834015

### PR DESCRIPTION
**Pull request by cortinico**
When generating Retrofit interfaces, we need to remove the leading
slash from the generated path. Having a leading slash will force
the path to be absolute for the host and will make URLs like:
https://petstore.swagger.io/v2/
impossible to call.

Fixes #49

### DIFF
--- a/plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
+++ b/plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
@@ -33,7 +33,7 @@ class KotlinGenerator : SharedCodegen() {
 
     private val apiDocPath = "docs/"
     private val modelDocPath = "docs/"
-    private var basePath: String? = null
+    internal var basePath: String? = null
 
     private val retrofitImport = mapOf(
             "GET" to "retrofit2.http.GET",
--- a/plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
+++ b/plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
@@ -33,6 +33,7 @@ class KotlinGenerator : SharedCodegen() {
 
     private val apiDocPath = "docs/"
     private val modelDocPath = "docs/"
+    private var basePath: String? = null
 
     private val retrofitImport = mapOf(
             "GET" to "retrofit2.http.GET",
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by cortinico** [comment]
@macisamuele What do you think about this PR?

**Comment by macisamuele** [comment]
@cortinico I'll check this PR in the next days. I'm not entirely sure that the approach is fully compatible with the swagger specifications (I need to check more deeply the related issue)

File: plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt [hunk 1]
```diff
@@ -407,6 +408,11 @@ class KotlinGenerator : SharedCodegen() {
         getHeadersToIgnore().forEach { headerName ->
             ignoreHeaderParameter(headerName, codegenOperation)
         }
+
+        // Let's remove the leading
+        if (!basePath.isNullOrBlank()) {
```
macisamuele: I did check few introductory docs and [videos](https://www.youtube.com/watch?v=KIAoQbAu3eA&feature=youtu.be&t=32m19s) related to this issue and the proposed approach looks the safest one to use in this case.

One not is related to the value of `base_path`.

If the spec specify `base_path` as `/<something_not_suffixed_by_slash>` (ie. `/v3` in the petstore APIs) how is retrofit going to build the final url?

cortinico: Given that we don't build the Retrofit base url (but is delegated to the user of the generator), I don't think this is an issue that we can address right now.

PR #52, opened 2019-07-26
Fix handling of Top Level Operation Headers

### LINKED ISSUE #45: Empty Headers added to the API  when there no operationId in Swagger
Empty header added to the API ( @Headers() ) when there no `operationId`. This is giving   runtime error 
`java.lang.IllegalArgumentException: @Headers annotation is empty.`

Remove the empty header or add  auto-generated operationId  like "X-Operation-ID: deviceGet" 

> Empty operationId found for path: get /devices. Renamed to auto-generated operationId: devicesGet

Ex: 

Bellow endpoint doesn't include `opertationID` which add empty `@Headers() ` while generating the code. 

```json
       "/devices": {
            "get": {
                "summary": "Fetch a list of Devices",
                "description": "Without any params, returns a list of the user's devices",
                "tags": [
                    "Device"
                ],
                "parameters": [
                    {
                        "$ref": "#/parameters/all"
                    },
                    {
                        "$ref": "#/parameters/userId"
                    },
                    {
                        "name": "id",
                        "in": "query",
                        "description": "To fetch one or more devices. Multiple params can be passed like `id=31&id=42`",
                        "required": false,
                        "type": "integer",
                        "collectionFormat": "multi"
                    },
                    {
                        "name": "uniqueId",
                        "in": "query",
                        "description": "To fetch one or more devices. Multiple params can be passed like `uniqueId=333331&uniqieId=44442`",
                        "required": false,
                        "type": "string",
                        "collectionFormat": "multi"
                    }
                ],
                "responses": {
                    "200": {
                        "description": "OK",
                        "schema": {
                            "type": "array",
                            "items": {
                                "$ref": "#/definitions/Device"
                            }
                        }
                    },
                    "400": {
                        "description": "No permission"
                    }
                }
            }
```

Screen Shot 2019-07-11 at 11 37 01 am

cortinico:
This PR cleans up the mustache template responsible of handling the generation of top level Retrofit headers.

Headers are now generated inside the KotlinGenerator class and passed to the template as a list of pair. This makes the handling of 0, 1, more header scenarios easier both on the generator and on the mustache template.

JUnit tests are attached.

Fixes #45

### DIFF
--- a/plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
+++ b/plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
@@ -28,7 +28,11 @@ class KotlinGenerator : SharedCodegen() {
          * This number represents the version of the kotlin template
          * Please note that is independent from the Plugin version
          */
-        @JvmStatic val VERSION = 11
+        @JvmStatic val VERSION = 12
+
+        // Vendor Extension use to generate the list of top level headers
+        const val HAS_OPERATION_HEADERS = "hasOperationHeaders"
+        const val OPERATION_HEADERS = "operationHeaders"
     }
 
     private val apiDocPath = "docs/"
@@ -404,7 +408,6 @@ class KotlinGenerator : SharedCodegen() {
 
         codegenOperation.imports.add("retrofit2.http.Headers")
         codegenOperation.vendorExtensions[X_OPERATION_ID] = operation?.operationId
-
         getHeadersToIgnore().forEach { headerName ->
             ignoreHeaderParameter(headerName, codegenOperation)
         }
@@ -413,6 +416,7 @@ class KotlinGenerator : SharedCodegen() {
         if (!basePath.isNullOrBlank()) {
             codegenOperation.path = codegenOperation.path.removePrefix("/")
         }
+        processTopLevelHeaders(codegenOperation)
         return codegenOperation
     }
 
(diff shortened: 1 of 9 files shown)


plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
```diff
@@ -451,4 +455,30 @@ class KotlinGenerator : SharedCodegen() {
...
+        val topLevelHeaders = mutableListOf<Pair<String, String>>()
+
+        // Send the X-Operation-Id header only if a custom operation ID was set.
+        val operationId = operation.vendorExtensions[X_OPERATION_ID] as String?
+        if (!operationId.isNullOrBlank()) {
+            topLevelHeaders.add(HEADER_X_OPERATION_ID to operationId)
+        }
+
+        // Send the Content-Type header for the first `consume` mediaType specified.
+        val firstContentType = operation.consumes?.firstOrNull()
```
macisamuele: `consumes` might be defined on two different places:
 * `#/consumes` [doc](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#swagger-object): default consume for all the operations
 * `#/paths/<http_path>/<http_verb>/consumes` [doc](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#operation-object): consume override for the specific operation

does `operation.consumes` honours the default value in case the `consumes` defined in the operation is missing? If not, then we should be dealing with such case as well

cortinico: > does operation.consumes honours the default value in case the consumes defined in the operation is missing? If not, then we should be dealing with such case as well

I don't think that the top level `consumes` should be translated to a `@Header` annotation. Ideally a top level `consumes` information should be treated by an OkHttp interceptor at the client level.

macisamuele: I manually verified the assumption and yes `CodegenOperation` does honours the top level `consumes` so this safe is good to go

PR #57, opened 2019-07-31
Update readme

doug-precocity:


cortinico:
Thanks for the PR @doug-precocity 
Apparently we actually have a bug in the setup of the project. I'm fixing it in #59 and I don't think we need the extra message in the README anymore.

PR #68, opened 2019-09-19
Add Support for moshi-codegen

### LINKED ISSUE #62: Add support for moshi-kotlin-codegen
We're currently generating code that relies on `moshi-kotlin`:

https://github.com/Yelp/swagger-gradle-codegen/blob/ed70186739e313de3a51e587d8b29b24da59b879/plugin/src/main/resources/kotlin/tools/GeneratedCodeConverters.kt.mustache#L10-L14

This has the [side effect](https://github.com/square/moshi#reflection) of pulling in the `kotlin-reflect` dependency that is a pretty big jar.

We should consider moving over to [moshi-kotlin-codegen](https://github.com/square/moshi#reflection).

Ideally we just need to generate the `@JsonClass` annotation for our data classes and make sure we provide proper proguard rules for it.

filipemp:
Fixes #62

### DIFF
--- a/plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
+++ b/plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
@@ -192,6 +192,7 @@ class KotlinGenerator : SharedCodegen() {
         // If there are any vars, we will mark them with the @Json annotation so we have to make sure to import it
         if (codegenModel.vars.isNotEmpty() || codegenModel.isEnum) {
             codegenModel.imports.add("com.squareup.moshi.Json")
+            codegenModel.imports.add("com.squareup.moshi.JsonClass")
         }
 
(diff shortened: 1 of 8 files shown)


cortinico:
Ping @filipemp 

Moshi 1.9 was released that should fix the problems you had.

filipemp:
@cortinico I just updated Moshi's version to see if it would work, locally it did. I'm not sure if we need proguard though or if not having is actually breaking the build. I'll try to get back to this over the next week or so

cortinico:
Awesome :)

> I'm not sure if we need proguard though or if not having is actually breaking the build

Ideally inspecting the artifact and checking if the `moshi-codegen` generated classes are there, should suffice.

filipemp:
It's gonna take awhile before I can actually get back to this, if anyone else wants to pick this up, it's free. Otherwise, when I'm free again I'll work on it

PR #69, opened 2019-09-20
Add Gradle lazy configuration

### LINKED ISSUE #63: Update the Gradle Plugin API to use Lazy Configuration
We should consider updating our Gradle plugin API to use the Gradle Lazy Configuration approach (`Provider` and `Property` to allow lazy configuration of the task).
https://docs.gradle.org/current/userguide/lazy_configuration.html

More context on this in the thread on #16 

Currently the `inputFile` property is the only required property to properly configure `swagger-gradle-codenge`. If the `inputFile` property is not provided, the gradle build is broken.
Ideally we would love to break just the execution of the `generateSwagger` task and provide an informative message to the user.

filipemp:
Takes advantage of
https://docs.gradle.org/current/userguide/lazy_configuration.html to
instantiate configuration of the plugin's task so that when configuring
the plugin it will only break if generateSwagger is run

Fixes #63


README.md
```diff
@@ -93,14 +93,14 @@ To configure the generator, please use the `generateSwagger { }` block. Here an
 
 ```kotlin
 generateSwagger {
-    platform = "kotlin"
```
cortinico: Can use the Groovy DSL here?

plugin/src/main/java/com/yelp/codegen/plugin/FeatureConfiguration.kt
```diff
@@ -7,5 +7,5 @@ open class FeatureConfiguration {
 
     @Input
     @Optional
-    var headersToRemove: Array<String> = arrayOf()
+    var headersToRemove: Array<String> = emptyArray()
```
cortinico: Shouldn't this be converted to a `Provider`/`Property` as well?

plugin/src/main/java/com/yelp/codegen/plugin/GenerateTask.kt
```diff
@@ -30,44 +33,51 @@ open class GenerateTask : DefaultTask() {
     @Input
     @Optional
     @Option(option = "platform", description = "Configures the platform that is used for generating the code.")
-    var platform: String? = null
+    var platformProvider: Provider<String> = project.objects.property(String::class.javaObjectType)
```
cortinico: Name should not be changed here as this is part of the public API (same for all of them).

plugin/src/main/java/com/yelp/codegen/plugin/GenerateTask.kt
```diff
@@ -116,23 +126,24 @@ open class GenerateTask : DefaultTask() {
...
         if (source != null && destin != null) {
             source.copyRecursively(destin, overwrite = true)
         }
     }
 
     private fun readVersionFromSpecfile(specFile: File) {
         val swaggerSpec = SwaggerParser().readWithInfo(specFile.absolutePath, listOf(), false).swagger
-        specVersion = when (val version = swaggerSpec.info.version) {
+
+        this.specVersionProvider = when (val version = swaggerSpec.info.version) {
```
cortinico: You probably don't need to re-assign `this.specVersionProvider` here

martinbonnin:
@filipemp are you still working on this ? I might pick up where you left.

filipemp:
> @filipemp are you still working on this ? I might pick up where you left.

Nope, go for it

PR #78, opened 2019-12-26
Fix Nullability warning in JsonAdapters

### LINKED ISSUE #74: Nullability change in moshi JsonAdapter
I'm getting errors on the generated classes for 
- LocalDateAdapter
- ZonedDateTimeAdapter
- BigDecimalJsonAdapter

The method 
```
override fun toJson(writer: JsonWriter?, value: ZonedDateTime?) {
```
seems to have changed to
```
override fun toJson(writer: JsonWriter, value: ZonedDateTime?) {
```
but I'm using the same moshi versions as the samples (1.8.0)

Screen Shot 2019-12-13 at 20 09 35

cortinico:
Updating generated methods `toJson` to accept a non null `JsonWriter`.
Previously the IDE was raising an issue due to different method signature.

Fixes #74

Ping @macisamuele for the review

### DIFF
--- a/plugin/src/main/resources/kotlin/tools/TypesAdapters.kt.mustache
+++ b/plugin/src/main/resources/kotlin/tools/TypesAdapters.kt.mustache
@@ -59,8 +59,8 @@ internal class LocalDateAdapter : XNullableJsonAdapter<LocalDate>() {
 
     override fun fromNonNullString(nextString: String) : LocalDate = LocalDate.parse(nextString, formatter)
 
-    override fun toJson(writer: JsonWriter?, value: LocalDate?) {
-        value?.let { writer?.value(it.format(formatter)) }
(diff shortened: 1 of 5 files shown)


macisamuele:
@cortinico please fix tests first.
Additionally this PR seems including some changes (operation id header) of #79 ... please make sure to make those changes only once ;)

cortinico:
ping @macisamuele for another pass of review (for some reasons I can't add you as reviewer)

cortinico:
> Do you think that we could have a test that can help us to avoid such 
> conditions?

So ideally we could:
- Remove the `generated-code` sample. The idea of this sample was to commit the generated code. Now that we have `junit-test`, the  former is probably not needed anymore and is yet another thing to maintain.
- Have a pre-commit hook that rebuilds everything before every commit.

PR #79, opened 2019-12-26
Fix missing inner enum import on operations

### LINKED ISSUE #76: Missing enum import
My swagger.json was generated by swashbuckle (from an asp.net core web service). I trying to generate a kotlin client from this swagger.json, the generation run successfully. But on the api generated object enum import is missing.

I followed this [sample code](https://github.com/Yelp/swagger-gradle-codegen/tree/master/samples/generated-code). Maybe I missing somethings.

You can reproduce the bug from this repository : https://github.com/midoriiro/yelp-swagger-gradle-codegen-missing-enum-import

cortinico:
Previously inner types of complex types were not imported properly on
operations. This resulted in Retrofit interfaces missing some imports.

This commit introduces methods to inspect and import the inner types
used as result types of operations.

Fixes #76

Ping @macisamuele for the review

### DIFF
--- a/plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
+++ b/plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
@@ -142,9 +142,25 @@ class KotlinGenerator : SharedCodegen() {
     public override fun listTypeWrapper(listType: String, innerType: String) =
             "$listType<$innerType>"
 
+    public override fun listTypeUnwrapper(baseType: String) = baseType
+            .replace(" ", "")
+            .removeSuffix(">")
+            .removePrefix("${typeMapping["list"]}<")
+
+    public override fun isListTypeWrapped(baseType: String) =
+            baseType.endsWith('>') && baseType.startsWith("${typeMapping["list"]}<")
+
     @VisibleForTesting
     public override fun mapTypeWrapper(mapType: String, innerType: String) =
             "$mapType<${typeMapping["string"]}, $innerType>"
+
+    public override fun mapTypeUnwrapper(baseType: String) = baseType
+            .replace(" ", "")
+            .removeSuffix(">")
+            .removePrefix("${typeMapping["map"]}<${typeMapping["string"]},")
+
+    public override fun isMapTypeWrapped(baseType: String) =
+            baseType.endsWith('>') && baseType.startsWith("${typeMapping["map"]}<${typeMapping["string"]},")
 
     @VisibleForTesting
     public override fun nullableTypeWrapper(baseType: String) =
@@ -407,6 +423,7 @@ class KotlinGenerator : SharedCodegen() {
         }
 
         codegenOperation.imports.add("retrofit2.http.Headers")
+
         codegenOperation.vendorExtensions[X_OPERATION_ID] = operation?.operationId
         getHeadersToIgnore().forEach { headerName ->
             ignoreHeaderParameter(headerName, codegenOperation)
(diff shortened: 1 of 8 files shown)


plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
```diff
@@ -142,10 +142,27 @@ class KotlinGenerator : SharedCodegen() {
     public override fun listTypeWrapper(listType: String, innerType: String) =
             "$listType<$innerType>"
 
+    public override fun listTypeUnwrapper(baseType: String) = baseType
+            .replace(" ", "")
+            .removeSuffix(">")
+            .removePrefix("List<")
+            .removePrefix("Array<")
```
macisamuele: As far as I can understand from the code the method will ever be called with `baseType` set to the type that will be output-ed on the file.
If the understanding is correct, then `Array<` should never be present

plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
```diff
@@ -142,10 +142,27 @@ class KotlinGenerator : SharedCodegen() {
     public override fun listTypeWrapper(listType: String, innerType: String) =
             "$listType<$innerType>"
 
+    public override fun listTypeUnwrapper(baseType: String) = baseType
+            .replace(" ", "")
+            .removeSuffix(">")
+            .removePrefix("List<")
```
macisamuele: I'd rather prefer using [`kotlinTypeMapping["list"]`](https://github.com/Yelp/swagger-gradle-codegen/blob/master/plugin/src/main/java/com/yelp/codegen/utils/KotlinLangUtils.kt#L121) so it might be moved into `SharedCodegen` as it will be _kotlin_ indpendent

plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
```diff
@@ -142,10 +142,27 @@ class KotlinGenerator : SharedCodegen() {
...
+            .removePrefix("Array<")
+
+    public override fun isListTypeWrapped(baseType: String) =
+            baseType.endsWith('>') && (baseType.startsWith("Array<") || baseType.startsWith("List<"))
```
macisamuele: As before: `Array` should not be possible and I would replace List with `kotlinTypeMapping["list"]`

plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
```diff
@@ -142,10 +142,27 @@ class KotlinGenerator : SharedCodegen() {
...
+
     @VisibleForTesting
     public override fun mapTypeWrapper(mapType: String, innerType: String) =
             "$mapType<${typeMapping["string"]}, $innerType>"
 
+    public override fun mapTypeUnwrapper(baseType: String) = baseType
+            .replace(" ", "")
+            .removeSuffix(">")
+            .removePrefix("Map<${typeMapping["string"]},")
```
macisamuele: ditto: `s/Map/typeMapping["map"]/`

plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
cortinico: Yup, you're right. Let's refactor this out :)

plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
cortinico: ++

cortinico:
@macisamuele ready for another pass :) I've addressed all the comments.

> PS: personally I would have replaced isMapTypeWrapped with a regex matching of 
> the mapTypeWrapper (but not blocking)

You're more than welcome to add it once this gets merged :)

PR #80, opened 2020-01-20
Update several dependencies

cortinico:
Getting ready for releasing `1.3.0` by updating all the dependencies

bdda4b0 MockWebServer to 3.12.6
0a92f08 RxJava2 to 2.2.17
8e360ea ThreeTenABP to 1.2.2
0d0c411 Retrofit to 2.7.1
5d7163f Moshi to 1.9.2
21d4913 com.gradle.plugin-publish to 0.10.1
1872432 Detetk to 1.4.0
b728a25 Gradle to 6.1
ed3eece AGP to 3.5.3
8ffaba9 Kotlin to 1.3.61


config/detekt/detekt.yml
```diff
@@ -0,0 +1,604 @@
+build:
+  maxIssues: 1
+  excludeCorrectable: false
+  weights:
+    # complexity: 2
+    # LongParameterList: 1
+    # style: 1
+    # comments: 1
```
macisamuele: nit. I don't think that these commented out configurations are needed.
I would cleanup this config file to only contain what you need.

**NOTE**:  Keeping the comments change the type `weights` from a map to a null value
The same might be applied to `#/processors/exclude`, `#/console-reports/exclude`, etc.

cortinico: I'd rather keep this file as it is as this is the output of `./gradlew plugin:detektGenerateConfig`.

I've just reapplied the customization we had on `plugin/detekt-config.yml` (now deleted).
Modifying this file will make diffing in the future harder.

macisamuele: 👍

PR #82, opened 2020-01-22
Allow Code-geneated Moshi Adapters

### LINKED ISSUE #62: Add support for moshi-kotlin-codegen

### LINKED ISSUE #71: Fix generated Kotlin enums for Moshi and Proguard/R8
When generating Kotlin enums, they should be annotated with 
`@JsonClass(generateAdapter = false)`
so that they still work with Moshi even when minification is enabled.
See https://github.com/square/moshi/issues/689

macisamuele:
The goal of this PR is pretty straight forward it plugs-in [`moshi-kotlin-codegen`](https://github.com/square/moshi#codegen) into the generated code.

The main difference in the code generated by swagger-gradle-codegen is the presence of `@JsonClass(generateAdapter = true)` within the code.

This annotation allows KotlinJsonAdapterFactory to [identify the JsonAdapter](https://github.com/square/moshi/blob/moshi-parent-1.9.2/kotlin/reflect/src/main/java/com/squareup/moshi/kotlin/reflect/KotlinJsonAdapter.kt#L191) generated via kotlin annotation processor (kapt) if the kapt directive (look into  `samples/junit-tests/build.gradle` changes) is enabled.

This approach leads to a no-op change for users of the current tool but empowers them to get better runtime performances by enabling kapt.
Disclaimer: I've not verified the performance improvements but considering that this is an opt-in approach I'm not concerned about it

Fixes #68 
Fixes #62 
Fixes #71


plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
```diff
@@ -205,6 +205,8 @@ open class KotlinGenerator : SharedCodegen() {
 
     @VisibleForTesting
     internal fun addRequiredImports(codegenModel: CodegenModel) {
+        // Import JsonClass annotation to enable Adapters generations via Kotlin Annotation Processor
+        codegenModel.imports.add("com.squareup.moshi.JsonClass")
```
cortinico: Can be moved inside the `if` block below

plugin/src/main/resources/kotlin/data_class.mustache
```diff
@@ -4,6 +4,7 @@
  * @property {{{name}}} {{description}}
 {{/vars}}
  */
+@JsonClass(generateAdapter = true)
```
cortinico: You probably need to add the same annotation on top of the `enum class` below

samples/kotlin-android-with-moshi-kotlin-codegen/build.gradle.kts
```diff
@@ -0,0 +1,56 @@
+plugins {
```
cortinico: Do we really need a new sample? We're switching the whole stack from Reflect to Codegen. The changes will be already reflected on the `kotlin-android`, `junit-tests`, `generated-code` samples

plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
macisamuele: I totally missed the other PRs that were opened ... sorry :(

I'll check if we can move the import into the if statement, I'll let you know soon about the progress.

plugin/src/main/resources/kotlin/data_class.mustache
macisamuele: For some reason I did not notice issues if I would not do it, I'll try to debug it a bit longer to see that the assumptions are verified and if we need to do it then we definitely need to create a test-scenario for it in order to make sure that we don't introduce a regression.

Side note: it's highly probable that I've not seen an issue as moshi-kotlin (and so reflection was still present)

cortinico:
Ideally if you could rebase your branch in top of Filipe's would be great
so we have his contribution in the graph as well.

samples/kotlin-android-with-moshi-kotlin-codegen/build.gradle.kts
macisamuele: The idea of this PR was mostly to allow both to work.

Ideally I'd like to have a version released that supports both approaches and then if we decide that generated-code is the way forward (the performance improvements are balanced by extra size, etc. ) than we'll advocate only for generated-code removing the reflection approach from the codebase

macisamuele:
Sure thing, as mentioned I did not see (I should have looked more carefully) PRs already trying to address the same issue.
As soon as I got some extra time I'll try to verify the presence of blockers and ensure that attribution is respected ;)

plugin/src/main/resources/kotlin/data_class.mustache
macisamuele: I've tried adding `@JsonClass(generateAdapter = true)` to the enum class and annotation processing fails with the following error `code gen for enums is not supported or necessary`.

Defining it as `@JsonClass(generateAdapter = false)` (as in #68) is doable but essentially no-op, so I would rather prefer not generating it at all.

Feel free to re-open it if you think that we need to annotate it anyway.

samples/kotlin-android-with-moshi-kotlin-codegen/build.gradle.kts
macisamuele: We did chat around this specific issue offline and we decided that:
 * for the next minor releases the plugin will support both reflective and codegen approaches
 * for the next major release we will be support only codegen approach by still preserving the capability of injecting `KotlinJsonAdapterFactory` (reflective approach).

We've decided that with this approach we can limit eventual issues that might surface by proguard configurations as well as gaining confidence around the final approach.

I'll be creating an issue (that will act as reminder) to fully remove the reflective approach

plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
macisamuele: `JsonClass` should be imported in all the model files, except if the model is actually a type alias.
Going to address this edge-case

macisamuele:
@cortinico, @filipemp: I did re-push the changes as I made the pre-commit hooks configuration less greedy ;)
Instead of ignoring the all samples directory I'm ignoring them into `samples/.*/src/main/java/`
NOTE: This is done because at the current state the codedgen does not really produce pretty source (according to the lint).
I will eventually look at it, but the reports are mostly related to wrong indentation and/or not needed imports.

cortinico:
Can we have the re-formatting + pre-commit change on a separate change? This makes checking out the PR Diffs for third party easier.

macisamuele:
@cortinico just to be sure, before publishing the changes, are you asking me to remove commits 
e835456 and dcf4e3d ?
If so I'm more than happy to remove it from the PR and publish a new PR once this is merged 😃

cortinico:
Yes correct. Let's keep the "infra" work outside of this PR.

.gitignore
```diff
@@ -5,3 +5,7 @@ java_pid*.hprof
 local.properties
 out/
 venv/
+
+# Samples folders -> content is generated by running ./gradlew generateSwagger
+samples/kotlin-android-with-moshi-kotlin-codegen/src/main/java/
+samples/kotlin-coroutines/src/main/java/
```
cortinico: We do have local `.gitignore` files for other samples so we should probably do the same for those two examples.

samples/kotlin-android/build.gradle.kts
```diff
@@ -12,6 +12,12 @@ android {
         versionCode = 1
         versionName = "1.0"
     }
+    buildTypes {
+        getByName("release") {
+            isMinifyEnabled = true
+            proguardFiles("moshi.pro")
+        }
+    }
```
cortinico: Why this? I don't think this is needed here (or at least I don't understand why you add it here and not to `groovy-android` or `kotlin-coroutines` as well).

settings.gradle.kts
```diff
@@ -18,6 +18,7 @@ pluginManagement {
 include(":plugin",
         ":samples:junit-tests",
         ":samples:kotlin-android",
+        ":samples:kotlin-android-with-moshi-kotlin-codegen",
```
cortinico: I'd rather have a `kotlin-android-moshi-codegen` just to keep it shorter

macisamuele: I've just find out that we're ignoring all the hidden files and so adding a gitignore file would have no effect.
I'm going to ignore specific hidden directories (`.idea` and `.gradle`) such that gitignore files can properly be traced by git.

samples/kotlin-android/build.gradle.kts
macisamuele: The config are not present on groovy-android because I forgot, while kotlin-coroutines did not made me thing to an android "project".

Still I think that is going to be simpler if we just provide one example of proguard rules.
As ideally the way forward will be only moshi-codegen I'm actually going to remove the definitions from kotlin-android

PR #86, opened 2020-01-27
Pre commit hooks and Samples update

macisamuele:
As mentioned in [#82](https://github.com/Yelp/swagger-gradle-codegen/pull/82#issuecomment-578424422) this follow-up PR ensures:
 * samples are updated with code on the branch
 * pre-commit are updated and all the repo is re-validated against it


.pre-commit-config.yaml
```diff
@@ -12,16 +12,20 @@ repos:
...
+    exclude: ^config/detekt/detekt\.yml$
 - repo: https://github.com/Yelp/detect-secrets
-  rev: v0.12.0
+  rev: v0.13.0
   hooks:
   - id: detect-secrets
     args: [--baseline, .secrets.baseline]
+
+# Do NOT run pre-commit hooks on generated-code
+exclude: ^samples/[^/]+/src/main/java
```
cortinico: Are we sure about this?
KtLint was sanitising the generated code a lot. Now it doesn't really look that elegant as before. Any arguments for having this `exclude` rule here?

samples/generated-code/src/main/java/.swagger-codegen-ignore
```diff
@@ -0,0 +1,23 @@
...
+
+# You can recursively match patterns against a directory, file or extension with a double asterisk (**):
+#foo/**/qux
+# This matches foo/bar/qux, foo/baz/qux, and foo/bar/baz/qux
+
+# You can also negate patterns with an exclamation (!).
+# For example, you can ignore all files in a docs folder with the file extension .md:
+#docs/*.md
+# Then explicitly reverse the ignore rule for a single file:
+#!docs/README.md
```
cortinico: This should not be committed. Why are you adding it?

samples/generated-code/src/main/java/.swagger-codegen/VERSION
```diff
@@ -0,0 +1 @@
+2.3.1
```
cortinico: Same as above.

.pre-commit-config.yaml
macisamuele: The proposal here, and I'm more than happy to remove this line if needed, is that `samples` should contain the exact output of the tool. 

If the generated code does not look "elegant" then we should eventually update the tool to generate some more elegant result rather than having ktlint fixing it.

One of the many things that I was noticing is that we have `import okhttp3.RequestBody` on all *Api.kt files even if we don't have parameters of `type: file`.

samples/generated-code/src/main/java/.swagger-codegen-ignore
macisamuele: As above ^

NOTE: If you feel like we should even generate this files we can suspend their generation, but as we do generate it we should have it in samples

.pre-commit-config.yaml
cortinico: > If the generated code does not look "elegant" then we should eventually update the tool to generate some more elegant result rather than having ktlint fixing it.

This is not always possible due to limitations of mustache. 
I'm generally fine with:

generator outputting not elegant code -> formatter fixing it -> eventually committing it.

samples/generated-code/src/main/java/.swagger-codegen-ignore
cortinico: As a side note, we should probably nuke `samples/generated-code/` as it's not really useful anymore and is yet another generated set of files that we have to handle.

.pre-commit-config.yaml
macisamuele: What might eventually be a concern is that having the formatter fixing the file might hide issues within the generated code.

Think about the [`import okhttp3.RequestBody`](https://github.com/Yelp/swagger-gradle-codegen/blob/9d9edcede7e2c719f9e41cec2ae9609444cd92ae/plugin/src/main/resources/kotlin/retrofit2/api.mustache#L4). The linter removes it because `RequestBody` is not used in most of the generated files. That's great, but what if we are using the generated code into a pipeline that generates an artifact? Probably we're going to have issues because the import statement (with the typo) will fail or similar.

About mustache limitations, I agree, I'll try to work on it to understand how complex would be to generate "elegant" enough code.

macisamuele:
In order to get the pre-commit hooks and samples updated I'm removing from this PR the pre-commit exclusion of the samples.

I'll address https://github.com/Yelp/swagger-gradle-codegen/pull/86#discussion_r371744767 into a subsequent PR (the new PR will mostly focus on having mustache generate a "nicer" code that will make ktlint a bit happier).

samples/generated-code/src/main/java/.swagger-codegen-ignore
macisamuele: Removed for now

.pre-commit-config.yaml
macisamuele: Keeping this open as we're going to move the effort to address this specific comment into a separate PR

cortinico:
> * ensure that pre-commit does not alter the code generated by the tool

Please update the description of this PR

PR #87, opened 2020-02-03
Add a simple plugin test

martinbonnin:
Keeping the ball rolling after FOSDEM :)

This pull request will help implementing and testing gradle lazy configuration (https://github.com/Yelp/swagger-gradle-codegen/issues/63)

It adds a very basic plugin test. The test just checks that the `generateSwagger` task is properly executed. More tests can be added later.


plugin/build.gradle.kts
```diff
@@ -11,6 +11,7 @@ plugins {
     kotlin("jvm") version "1.3.61"
     id("com.gradle.plugin-publish") version "0.10.1"
     id("io.gitlab.arturbosch.detekt") version "1.4.0"
+    id("maven-publish")
```
martinbonnin: This is unfortunately needed to deploy to a local repository. More below.

plugin/build.gradle.kts
```diff
@@ -69,3 +107,7 @@ tasks.jacocoTestReport {
 tasks.check {
     dependsOn(tasks.jacocoTestReport)
 }
+
+tasks.withType<Test> {
+    dependsOn("publishDefaultPublicationToPluginTestRepository")
```
martinbonnin: This is the main trick to test the plugin. It'll deploy the plugin artifacts to `rootDir/build/localMaven` so that the testProject can use this one and not the one on the gadle portal.

plugin/src/test/testProject/build.gradle.kts
```diff
@@ -0,0 +1,26 @@
+import com.yelp.codegen.plugin.GenerateTaskConfiguration
+
+buildscript {
+  repositories {
+    maven {
+      url = uri("../../../build/localMaven")
```
martinbonnin: This tells the testProject to use the local plugin.

plugin/build.gradle.kts
cortinico: You can safely replace with:

```
tasks.withType<Test> {
    dependsOn("publishToMavenLocal")
}
```

And use MavenLocal for the whole testing scenario.

plugin/build.gradle.kts
cortinico: There is actually already `maven-publish` applied just a couple of lines above :)

plugin/build.gradle.kts
```diff
@@ -53,6 +54,43 @@ pluginBundle {
     }
 }
 
+configure<PublishingExtension> {
```
cortinico: You can remove this `PublishingExtension` completely as it's not needed.

plugin/src/test/testProject/build.gradle.kts
```diff
@@ -0,0 +1,26 @@
+import com.yelp.codegen.plugin.GenerateTaskConfiguration
+
+buildscript {
+  repositories {
+    maven {
```
cortinico: You should be able to replace with `mavenLocal()` and it will work out of the box.

plugin/src/test/testProject/build.gradle.kts
```diff
@@ -0,0 +1,26 @@
...
+      url = uri("../../../build/localMaven")
+    }
+    jcenter()
```
cortinico: If you change this into:

```kotlin
    jcenter {
      content {
        excludeGroup("com.yelp.codegen")
      }
    }
```

we will be sure we're actually getting the plugin only from MavenLocal

plugin/src/test/testProject/settings.gradle.kts
```diff
@@ -0,0 +1 @@
+rootProject.name="testProject"
```
cortinico: Commenting here but is more of a general issue.
Please rename this directory to ` plugin/src/test/resources/` (or ` plugin/src/test/resources/testProject/`).

The reason behind this is that the `test` gradle task has the resources directory as InputDirectory. This means that a change of a file inside this directory (say `build.gradle.kts`) will trigger a cache invalidation and the task will be actually executed. With the current setup, running twice the `test` task will result in the task being skipped the second time, even if you edit the `build.gradle.kts` in between.

plugin/build.gradle.kts
martinbonnin: I don't like that too much since `mavenLocal` is global to the machine and might contain other artifacts that will be pulled in tests inadvertently. Doing everything inside the repo makes sure everything is contained.

plugin/build.gradle.kts
martinbonnin: Ooops, sorry, will remove

plugin/build.gradle.kts
martinbonnin: Not sure I follow. I could replace `configure<PublishingExtension> {}` with `publishing {}` to leverage generated accessors. I've been avoiding this lately though as I had the impression that intelliJ performed better without the generated accessors. 
Happy to change to `publishing {}` though as I don't really have any data to back this impression.

plugin/src/test/testProject/build.gradle.kts
martinbonnin: Nice ! Will do :+1:

plugin/src/test/testProject/settings.gradle.kts
martinbonnin: Good catch, thanks.

plugin/build.gradle.kts
cortinico: > Doing everything inside the repo makes sure everything is contained.

I see your point, but then having this in the test gradle file would probably suffice as well:
 
```
    mavenLocal {
      content {
        includeGroup("com.yelp.codegen")
      }
    }
```

This will make sure mavenLocal is used only for `com.yelp.codegen` and nothing else.

plugin/src/test/testProject/settings.gradle.kts
martinbonnin: I actually went a slightly different route and added `inputs.dir("src/test/testProject")` to the "test" task (See https://github.com/martinbonnin/swagger-gradle-codegen/blob/f8385348e283509e8cdc4bc03ab04b651bec857f/plugin/build.gradle.kts#L112)

The rationale is that the `testProject` files don't actually need to be resources, they're never accessed as such so it would be a bit misleading to put them in the resources directory. Let me know if it's ok for you.

plugin/build.gradle.kts
martinbonnin: Ah, now I got it! The `publication` block isn't required as the `com.gradle.plugin-publish` plugin already creates a `pluginMaven` publication. I'll remove it.

I kept the `repositories` block still as I feel using `mavenLocal` could be a bit dangerous (see above).

cortinico: > could be a bit dangerous (see above).

How do you feel about https://github.com/Yelp/swagger-gradle-codegen/pull/87#discussion_r374580723 ?

plugin/build.gradle.kts
martinbonnin: But then there's the other side of the thing. This will add the plugin under test plugin to mavenLocal, which a user might end up a few weeks later when wanting to pull something else from mavenLocal.

Overall, I find adding 

```
configure<PublishingExtension> {
    repositories {
        maven {
            name = "pluginTest"
            url = uri("file://${rootProject.buildDir}/localMaven")
        }
    }
}
```
is a ok price to pay for proper isolation.

plugin/src/test/testProject/settings.gradle.kts
cortinico: Yup that works for me

plugin/src/test/testProject/build.gradle.kts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+apply(plugin = "com.yelp.codegen.plugin")
+
+configure<GenerateTaskConfiguration> {
+  platform = "kotlin"
+  packageName = "com.yelp.codegen.samples"
+  // this file is in rootDir/plugin/build/testProject/
+  inputFile = file("../../../samples/junit-tests/junit_tests_specs.json")
+  outputDir = file("./build/generetadSources")
```
cortinico:
```suggestion
  outputDir = file("./build/generatedSources")
```

plugin/src/test/java/com/yelp/plugin/PluginTests.kt
```diff
@@ -0,0 +1,28 @@
...
+import org.gradle.testkit.runner.GradleRunner
+import org.gradle.testkit.runner.TaskOutcome
+import org.junit.Assert
+import org.junit.Test
+import java.io.File
+
+
+class PluginTests {
+  @Test
+  fun `basic plugin test`() {
```
cortinico:
```suggestion
  fun basicPluginTest() {
```

Let's stay consistent with the other tests 👍

plugin/build.gradle.kts
```diff
@@ -53,6 +53,15 @@ pluginBundle {
...
+    repositories {
+        maven {
+            name = "pluginTest"
+            url = uri("file://${rootProject.buildDir}/localMaven")
+        }
+    }
```
cortinico: Just leave a comment above this block to explain that we're publishing inside the `pluginTest` maven repo for test purposes. Otherwise could look harder to understand for external contributors.

cortinico:
Awesome first contrib 💪

PR #88, opened 2020-02-03
Annotate with @JsonClass classes with no parameters/variables

macisamuele:
This is needed because if we would fully remove `moshi-kotlin` dependency (as planned by version 2 release) we would not be able to properly have the Json to Object conversion and vice-versa.

The most important change introduced by this PR is visible into [`samples/junit-tests/.../models/EmptyModel.kt`](https://github.com/Yelp/swagger-gradle-codegen/compare/master...macisamuele:maci-support-moshi-codegen-empty-class-edge-case#diff-497b242f6dee935e9766b2969ed52f43).

The reason for forcing the presence of `@JsonClass(generateAdapter = true)` even if there are no attributes is related to the fact that if we would remove the usage of `KotlinJsonAdapterFactory` then we would not be able to covert the `EmptyModel` instance from/to `JSON`.

In order to validate it I've run [this script](https://gist.github.com/macisamuele/8705010811d9a20c2b12fd4a7372a049#file-script-sh).
The script basically ensures that we have no moshi-kotlin/kotlin-reflect dependencies on samples/junit-tests and runs the tests

The results (raw results in [here](https://gist.github.com/macisamuele/8705010811d9a20c2b12fd4a7372a049#file-script-sh-execution)), as expected, shows that:
 * **PR branch** leads to green tests
 * **current master** fails to run `emptyEndpointTest_withEmptyBody` test.
    The test failure contains the following exception (the raw exception is available [here](https://gist.github.com/macisamuele/8705010811d9a20c2b12fd4a7372a049#file-integral-exception))
```
java.lang.IllegalArgumentException: Unable to create converter for class com.yelp.codegen.generatecodesamples.models.EmptyModel
...
Caused by: java.lang.IllegalArgumentException: Cannot serialize Kotlin type com.yelp.codegen.generatecodesamples.models.EmptyModel. Reflective serialization of Kotlin classes without using kotlin-reflect has undefined and unexpected behavior. Please use KotlinJsonAdapter from the moshi-kotlin artifact or use code gen from the moshi-kotlin-codegen artifact.
```

### DIFF
--- a/plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
+++ b/plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
@@ -205,8 +205,13 @@ open class KotlinGenerator : SharedCodegen() {
     @VisibleForTesting
     internal fun addRequiredImports(codegenModel: CodegenModel) {
         // If there are any vars, we will mark them with the @Json annotation so we have to make sure to import it
-        if (codegenModel.vars.isNotEmpty() || codegenModel.isEnum) {
+        if (codegenModel.allVars.isNotEmpty() || codegenModel.isEnum) {
             codegenModel.imports.add("com.squareup.moshi.Json")
+        }
+
+        if (!codegenModel.isAlias) {
+            // If we are rendering a model (or enum) we are annotating it with @JsonClass,
+            // so we need to make sure that we're importing it
             codegenModel.imports.add("com.squareup.moshi.JsonClass")
         }
 
--- a/plugin/src/main/resources/kotlin/data_class.mustache
+++ b/plugin/src/main/resources/kotlin/data_class.mustache
@@ -4,8 +4,8 @@
  * @property {{{name}}} {{description}}
 {{/vars}}
  */
-{{#hasVars}}@JsonClass(generateAdapter = true)
-data {{/hasVars}}class {{classname}} {{#hasVars}}(
+@JsonClass(generateAdapter = true)
+{{#hasVars}}data {{/hasVars}}class {{classname}} {{#hasVars}}(
 {{#requiredVars}}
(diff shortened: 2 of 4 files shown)


plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
```diff
@@ -205,8 +205,13 @@ open class KotlinGenerator : SharedCodegen() {
     @VisibleForTesting
     internal fun addRequiredImports(codegenModel: CodegenModel) {
```
cortinico: Please make sure you write tests for this method. There are already some inside the KotlinGeneratorTest.kt just add cases for the `JsonClass`. 

Also adding a test would document why you changed from `vars` to `allVars`

macisamuele: The change of `codegenModel.vars.isNotEmpty()` to `codegenModel.allVars.isNotEmpty()` has been done to make the code more consistent as in line 219 we're iterating over `codegenModel.allVars`.

`vars` and `allVars` are equivalent while `supportsInheritance` is set to `false` (current state of the library).
>      // By default these are the same collections. Where the code generator supports inheritance, composed models
>      // store the complete closure of owned and inherited properties in allVars and allMandatory.
>      allVars = vars;

cortinico: Gotcha. The test request is still valid anyway.

macisamuele: Sure ;)
They are there already.

Just wanted to briefly mention why I've modified it ;)

cortinico:
LGTM!

PR #89, opened 2020-02-03
Remove samples/generated-code as it is duplicated by junit-tests

macisamuele:
This PR is created as result of https://github.com/Yelp/swagger-gradle-codegen/pull/86 conversation.

The generated-code samples are not much different than groovy-android sample except for the fact that it did contain the generated code. Unfortunately, we were not building, testing or more in general doing something with the code so it was there mostly as a sort showcase of the generator output.

As junit-tests does actually run tests on the generated code, and does include the plugin generated code we might just remove the `samples/generated-code` by hinting (on the README) the presence of `samples/junit-tests` as an example of the generated code.

Additionally to that the junit-tests sample does provide a possible implementation of integration tests that allow to verify the behaviour of the APIs and Models.


README.md
```diff
@@ -87,6 +85,7 @@ You can find some **examples** in this repository to help you set up your genera
 * [samples/kotlin-coroutines](/samples/kotlin-coroutines)<sup>*</sup> Contains an example of an Android Library configured to output Kotlin Coroutines capable code.
 
 * [samples/junit-tests](/samples/junit-tests) This sample contains specs used to test edge cases and scenarios that have been reported in the issue tracker or that are worth testing. Tests are executed using [`moshi-codegen`](https://github.com/square/moshi#codegen).
+ It does also contains all the generated code inside the `/scr/main/java` folder. You can use this example to see how the generated code will _look like_ (like [here](/samples/junit-tests/src/main/java/com/yelp/codegen/generatecodesamples/apis/ResourceApi.kt)).
```
cortinico: I think you need to add an extra space before this line

PR #90, opened 2020-02-04
Update mustache style

macisamuele:
The goal of this PR is to improve our mustache templates to allow generation of code that will pass untouched through our linters (ktlint will consider the code good-looking).

The PR introduces changes on:
 * generator code: to ensure that we're running actions (ie. version definition and stripping headers) in the shared codegen as it is not android specific and to ensure that we have the correct file imports (for android)
 * generator code constants: ie `Map<K,V>` should have a space in order to have ktlint happy with it
 * templates: to ensure that the correct new lines, spaces and similars are placed.

**I do I know it works as we'd like to?**
Changes into `samples/*` are essentially not present (if we exclude the removal of one indentation level of `X-Operation-ID` header).
Running
```
make regenerate-samples && pre-commit run --all-files && git status
```
returns an empty diff, so pre-commit considers the generated code "good-looking"


plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt
```diff
@@ -402,14 +402,20 @@ open class KotlinGenerator : SharedCodegen() {
...
                 if (importName !in codegenOperation.imports) {
                     codegenOperation.imports.add(importName)
                 }
             }
+
+            if (codegenParameter.isFile) {
+                codegenOperation.imports.add("okhttp3.RequestBody")
+                // The generated Retrofit APIs use RequestBody and not File objects
+                codegenOperation.imports.remove("File")
+            }
```
cortinico: I understand that you're wrapping the `"okhttp3.RequestBody"` here as otherwise you'll have a bunch of unused imports. On the other hand you're adding a new endpoint to `junit-test` that is not covered by a test. I'd suggest you add the new endpoint in a separate PR. Ideally we should tackle also the Multipart scenario that I'm pretty sure we don't cover.

macisamuele: I'd rather keep the already existing endpoint, I'll try to see how expensive would be creating a test for it.

macisamuele: Creating a realistic test for it seems going a bit beyond the scope of the PR, especially considering that I'm getting the feeling that files are not properly handled (#99 )

Going to offload the effort in a follow-up PR

PR #95, opened 2020-02-04
Use TemporaryFolder JUnit-Test rule to create temporary directories

macisamuele:
The goal of this PR is to unify the creation of temporary directories of tests under the same approach (TemporaryFolder junit rule).

### DIFF
--- a/plugin/src/test/java/com/yelp/codegen/MainTest.kt
+++ b/plugin/src/test/java/com/yelp/codegen/MainTest.kt
@@ -1,34 +1,32 @@
 package com.yelp.codegen
 import java.io.File
+import org.junit.Rule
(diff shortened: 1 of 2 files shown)


plugin/src/test/java/com/yelp/codegen/MainTest.kt
```diff
@@ -1,34 +1,32 @@
 package com.yelp.codegen
 import java.io.File
+import org.junit.Rule
 import org.junit.Test
 import org.junit.rules.TemporaryFolder
 
 class MainTest {
+    @Rule
+    @JvmField
```
cortinico: -> `@get:Rule`

PR #97, opened 2020-02-05
Make the root project a composite build

martinbonnin:
One more step towards lazy task configuration (https://github.com/Yelp/swagger-gradle-codegen/pull/69)

This pull requests ~moves the samples to a [composite build](https://docs.gradle.org/current/userguide/composite_builds.html)~ makes the rootProject a composite build that will treat the plugin as an included build.

This allows to do [dependency substitution](https://docs.gradle.org/current/userguide/resolution_rules.html#sec:dependency_substitution_rules) on the plugin artifact and therefore have the sample code in sync with the plugin code from the repo (without having to publish on ~mavenCentral~ mavenLocal or jcenter first).

~If you're working on the samples, you should open `swagger-gradle-codegen/sample` in your favorite IDE.~ Opening the root project in intelliJ is still the recommended way to work on the project as it will include the necessary projects.


martinbonnin:
I just realized there's more to it as we need to compile samples from the CI as well obviously. Before I dive into this, is this a path worth exploring or should we continue using `mavenLocal` ?

cortinico:
> This allows to do dependency substitution on the plugin artifact and therefore have the sample code in sync with the plugin code from the repo (without having to publish on mavenCentral or jcenter first).

You don't need to publish on `mavenCentral` or `jcenter` at the moment. You have to publish the plugin to `mavenLocal` and use it in a subsequent build.

> If you're working on the samples, you should open swagger-gradle-codegen/sample in your favorite IDE.

Ideally we want to avoid having to open two separate project to work with samples and/or plugins. The convenience of the current setup is that you open a project and you have everything there.

martinbonnin:
>> This allows to do dependency substitution on the plugin artifact and therefore have the sample code in sync with the plugin code from the repo (without having to publish on mavenCentral or jcenter first).
>
>You don't need to publish on mavenCentral or jcenter at the moment. You have to publish the plugin to mavenLocal and use it in a subsequent build.

Sorry, I meant mavenLocal, not mavenCentral :man_facepalming: . As I said in the [other PR](https://github.com/Yelp/swagger-gradle-codegen/pull/87), I'm not a fan of using mavenLocal as it introduces a global state that can have side effects and that needs to be maintained.

> Ideally we want to avoid having to open two separate project to work with samples and/or plugins.

Technically, you can do all the work from the composite build and edit the plugin files from there.

> The convenience of the current setup is that you open a project and you have everything there.

The drawback being that changing the plugin requires to call `publishToMavenLocal` for the changes to be seen by the samples. Also when the project is first opened, the samples use the plugin from jcenter and not the one from the repo, which might be confusing if they don't behave exactly the same.

Let me know what you think. I started to work on this while working on lazy configuration because I changed some API in `GenerateTaskConfiguration` and this broke my gradle sync. But now I know about `publishToMavenLocal`, I can do the lazy configuration without having composite builds.

cortinico:
Sounds like composite build could help us improve the project setup 👍 I'm in for it. I just ask you to unwrap the setup that you're proposing here.

Ideally the base project should be the one in the root folder. From there you should include all the sample modules (as gradle module) and `plugin` as a composite build.

In other words: in the top level `settings.gradle.kts` the goal is to have something like.

```kotlin
include(":samples:junit-tests",
        ":samples:kotlin-android",
        ":samples:kotlin-android-moshi-codegen",
        ":samples:kotlin-coroutines",
        ":samples:groovy-android")
includeBuild(":plugin")
```

This will make sure that by running `gradle` in the top folder, everything works as expected.

Makefile
```diff
@@ -6,20 +6,14 @@
 install-hooks: .git/hooks/pre-commit
 	@true
 
-regenerate-samples:
-	${CURDIR}/gradlew plugin:build
-	${CURDIR}/gradlew plugin:publishToMavenLocal
-	${CURDIR}/gradlew generateSwagger
-
```
martinbonnin: This is the main benefit from composite builds. I think this target isn't needed anymore since samples will rebuild the plugin on demand.

gradle.properties
```diff
@@ -1 +0,0 @@
-org.gradle.jvmargs=-Xmx4g -XX:MaxMetaspaceSize=4g
```
martinbonnin: I removed in favor of passing the parameter only when needed (the 'check' task). Please tell me if it's needed somewhere else.

settings.gradle.kts
```diff
@@ -15,9 +16,11 @@ pluginManagement {
...
 
-include(":plugin",
+include(
         ":samples:junit-tests",
         ":samples:kotlin-android",
         ":samples:kotlin-android-moshi-codegen",
         ":samples:kotlin-coroutines",
         ":samples:groovy-android")
+
+includeBuild("plugin-root")
```
martinbonnin: I ended up with `plugin-root` to make a distinction with the `plugin` module. Also if both project have the same name, it confuses the dependency substitution.

cortinico: Can I ask you to rename it to `gradle-plugin`? As `plugin` (or `plugin-root`) is less informative in this context?

plugin-root/settings.gradle.kts
```diff
@@ -0,0 +1,17 @@
...
+        jcenter()
+        google()
+    }
+    resolutionStrategy {
+        eachPlugin {
+            if ("com.android" in requested.id.id) {
+                useModule("com.android.tools.build:gradle:${requested.version}")
+            }
+        }
+    }
```
cortinico: This is not needed anymore. This `plugin` project doesn't have any `com.android` deps.

Makefile
```diff
@@ -6,20 +6,14 @@
...
 run-hooks: venv
 	${CURDIR}/venv/bin/pre-commit run --all-files
 
 test: run-hooks
-	${CURDIR}/gradlew plugin:build
-	${CURDIR}/gradlew plugin:publishToMavenLocal
+	${CURDIR}/gradlew -p plugin-root build check
```
cortinico: `-p` -> `--project-dir`
Easier to maintain and to understand

Makefile
```diff
@@ -6,20 +6,14 @@
...
 	${CURDIR}/gradlew generateSwagger
-	${CURDIR}/gradlew assembleDebug
-	${CURDIR}/gradlew check
+	# The check task requires a lot of MetaSpace
```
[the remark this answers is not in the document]
martinbonnin: Absolutely no idea. Maybe it's a leak, maybe it's something else. It was a global setting before this pull request. See https://github.com/Yelp/swagger-gradle-codegen/pull/97#pullrequestreview-354359623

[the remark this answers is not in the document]
cortinico: I assume we can remove this flag at all, and eventually re-add it if we see failures. This could be a copy-n-paste bug from somewhere else.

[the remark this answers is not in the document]
martinbonnin: It does fail without. For an example: https://travis-ci.com/Yelp/swagger-gradle-codegen/builds/147559551

plugin-root/gradle
```diff
@@ -0,0 +1 @@
+../gradle
```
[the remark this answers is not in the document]
martinbonnin: I removed this symlink as well, it seems to work without. I added them in case someone wanted to open `gradle-plugin`  from intelliJ but now that the root project is the composite one, it shouldn't matter much.

Makefile
```diff
@@ -6,20 +6,14 @@
 install-hooks: .git/hooks/pre-commit
 	@true
 
-regenerate-samples:
```
cortinico: Please don't remove this target as it's handy when regenerating all the samples.

martinbonnin: With the composite build, you should be able to just run `./gradlew generateSwagger` and it should regenerate the samples.
That's actually a more general question I had: why not making everything in gradle instead of `Makefile`? It feels a bit awkward to have two build systems.

cortinico: @macisamuele 👆

cortinico: That's a concern I already raised with @macisamuele. I'm also in for having composite build tasks rather than Gradle + Make. I generally don't have a really strong opinion on this and I'm fine with whatever makes everyone comfortable :)

macisamuele:
@martinbonnin I think that #101 caused merge conflicts with this branch. Sorry 😢 

Let me know if you need support or you are able to address them by yourself (I don't expect it being very complex)

martinbonnin:
@macisamuele this branch renames a lot of files so I guess that was expected. I just rebased on master, it should be good now.

macisamuele:
@martinbonnin ++
Looked like you had @cortinico approval for your changes. So I'll prefer having him taking care of merging this change.
From a quick pass through the changes I see why the conflicts were there and how rebasing fixed all the things, so it should be all good ™️ 

Thanks a lot for helping in improving our gradle configurations 💯

Makefile
cortinico: Let's merge this but keep the Make+Gradle discussion opened. I would love to get @macisamuele 's opinion on this.

macisamuele:
> keep the Make+Gradle discussion opened

@cortinico @martinbonnin : I'm the kind of lazy developer ;)
I do usually configure, in each git repo I'm working on, a custom git command (ie. `git push-after-tests`) that allows me to push to `origin` only if tests are green.
Said so I don't have a very strong opinion on having Make or Gradle or both, is more about having a single command to run that basically runs all the tests that travis would run.

Having a gradle task that would behave like the current `make test` would be more than enough on my side 😄

Makefile
macisamuele: Replacing `make regenerate-samples` with `./gradlew generateSwagger` looks like a very good to me.
I'll just, in a future PR, remove the  `regenerate-samples` `PHONY` (first Makefile line) 😅

cortinico:
I'm totally in for nuking the Makefile. @martinbonnin do you want to take this?

martinbonnin:
👍 I'll open a PR. What's a good name for the gradle task that'll do everything ? test is already taken. Something like `./gradlew buildProject` ? `./gradlew makeAll` ? Something else ?

macisamuele:
What about `travisTest`? At the end the task should be connected to the travis execution as well :blush:

cortinico:
I'd rather go for `preMerge` or `preMergeTest` as is CI agnostic and contains all the verification tasks that needs to be executed before integrating a branch

Issue #99, opened 2020-02-05
Verify correctness of endpoints with file parameters


macisamuele:
While working on #90 I was having troubles on writing a simple test that sends a file to the mock server.
This was mostly caused by the fact that the mock server is receiving the `toString` representation of the `RequestBody` and not its real content.

At this point I'm not 100% sure that the current plugin is able to generate proper retrofit APIs that allow file uploads.

The scope of this issue is to track the effort needed, if any, to ensure file upload capabilities.

macisamuele:
I do have some conclusive results over here.
I'm currently preparing a PR to address the issue and to write the missing tests of #90 .

According to the swagger specifications in order to send files the operation must consume `multipart/form-data` ([doc](https://swagger.io/docs/specification/2-0/file-upload/)). The junit-tests [file endpoint](https://github.com/Yelp/swagger-gradle-codegen/blob/7ae0186fb9dfd9e51fcf32a8ff56b3529811d90d/samples/junit-tests/junit_tests_specs.json#L576-L596) does not define `multipart/form-data` and so the generate code does not work well.

Considering this we have two alternatives: 
 1. Find a way to make it work anyway
 2. Stick to the swagger specifications and annotate the endpoint as unsafe so we will discourage usage of the endpoint.

I'm aiming to use the second alternative

PR #101, opened 2020-02-06
File uploads tests

macisamuele:
This PR has 2 main goals
 1. endpoint that are not multipart and that have file parameter should be consider unsafe
 2. add integration tests around the FileAPI 
Fixes: #99

In order to be sure that the "backend" will receive valid informations I've defined a `Dispatcher` that will returns a response containing details of the body:
 * is it [RFC2387](https://tools.ietf.org/html/rfc2387) compatible?
 * what is the boundary?
 * what are the details of the different body parts?


macisamuele:
@cortinico I'm going to publish a new iteration.
The reason for this are:
 * simplifying a bit the logic for extracting details of the request body
 * moving it into extension functions of `PostFileEndpointTest` instead of polluting `MockServerApiRule` (it was not really a great idea)
 * removing not needed objects on the specs

samples/junit-tests/src/test/java/com/yelp/codegen/generatecodesamples/PostFileEndpointTest.kt
```diff
@@ -0,0 +1,179 @@
...
+    val contentDisposition: String? = null,
+    val contentLength: String? = null,
+    val contentType: String? = null
+)
+
+private fun processPart(part: String): MultiPartInfo {
+
+    val partLines = part.split("\r\n").let {
+        // part would look like "\r\nHeader-Name: Header-Value\r\n...\r\n\r\nbody\r\n"
+        // Here we're removing the first and last part as we know that are empty and
```
cortinico: and...? 😅

samples/junit-tests/src/test/java/com/yelp/codegen/generatecodesamples/PostFileEndpointTest.kt
```diff
@@ -0,0 +1,179 @@
...
+        it.subList(1, it.size - 1)
```
cortinico: -> `.drop(1).dropLast(1)`

samples/junit-tests/src/test/java/com/yelp/codegen/generatecodesamples/PostFileEndpointTest.kt
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    val partLinesIterator = partLines.iterator()
+    val headers = mutableMapOf<String, String>()
```
cortinico:
```kotlin
val headers = part.split("\r\n")
            .drop(1)
            .dropLast(1)
            .filter { it.isNotEmpty() }
            .map { it.split(": ", limit = 2) }
            .associateBy({it[0]}, {it[1]})
```

samples/junit-tests/src/test/java/com/yelp/codegen/generatecodesamples/PostFileEndpointTest.kt
```diff
@@ -0,0 +1,179 @@
...
+import org.junit.Assert.assertTrue
+import org.junit.Rule
+import org.junit.Test
+
+private data class FileApiStats(
+    var boundary: String? = null,
+    var errors: List<String>? = null,
+    var parts: List<MultiPartInfo>? = null
+)
+private data class MultiPartInfo(
```
cortinico: add a space above

samples/junit-tests/src/test/java/com/yelp/codegen/generatecodesamples/PostFileEndpointTest.kt
```diff
@@ -0,0 +1,179 @@
...
+import okhttp3.RequestBody
+import okhttp3.mockwebserver.MockResponse
+import okhttp3.mockwebserver.RecordedRequest
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertTrue
+import org.junit.Rule
+import org.junit.Test
+
+private data class FileApiStats(
```
cortinico: Those util classes should go to a `MultiPartTestUtil.kt` file

samples/junit-tests/src/test/java/com/yelp/codegen/generatecodesamples/PostFileEndpointTest.kt
```diff
@@ -0,0 +1,179 @@
...
+    for (headerLine in partLinesIterator) {
+        if (headerLine.isEmpty()) {
+            break
+        } else {
+            val (name, value) = headerLine.split(": ", limit = 2)
+            headers[name] = value
+        }
+    }
+
+    val content = partLinesIterator.asSequence().joinToString("\r\n")
```
cortinico:
```suggestion
val content = part.split("\r\n")
            .drop(1)
            .dropLast(1)
            .joinToString("\r\n")
```

samples/junit-tests/src/test/java/com/yelp/codegen/generatecodesamples/PostFileEndpointTest.kt
```diff
@@ -0,0 +1,179 @@
...
+         *      <Content# Header Name>: <Content# Header Value>\r\n
+         *      [<Content# Header Name>: <Content# Header Value>\r\n]
+         *      \r\n
+         *      <Content#>\r\n
+         *      --<BOUNDARY>--\r\n
+         *      \r\n
+         */
+        val requestBody = this.body.readUtf8()
+
+        val boundaryLine = requestBody.splitToSequence("\r\n").first()
```
cortinico:
```suggestion
        val boundaryLine = requestBody.split("\r\n").first()
```

samples/junit-tests/src/test/java/com/yelp/codegen/generatecodesamples/PostFileEndpointTest.kt
```diff
@@ -0,0 +1,179 @@
...
+            errors.add("Boundary should start with '--'")
+        } else {
+            response.boundary = boundaryLine.substring(startIndex = 2)
+            bodyParts.addAll(requestBody.split(boundaryLine))
+            if (!bodyParts.last().startsWith("--")) {
+                errors.add("Start and End Boundaries are different")
+            } else if (bodyParts.last().substring(startIndex = 2) != "\r\n") {
+                errors.add("The last line of the body has to be empty")
+            }
+        }
```
cortinico: I'd remove the whole `errors` field at all and just call `fail()` here.
It's pointless to collect errors if you know that you want to let the test fail as soon as you have one.

samples/junit-tests/src/test/java/com/yelp/codegen/generatecodesamples/PostFileEndpointTest.kt
macisamuele: and ... and should not be there ;)

samples/junit-tests/src/test/java/com/yelp/codegen/generatecodesamples/PostFileEndpointTest.kt
macisamuele: I was using the iterator to traverse the all input only one.
It's kinda sad that kotlin (at least for what I found) does not allow to scan an iterator while consuming it

The subsequent variable `content` represent the content of the file.

Considering this I would prefer using the _iterator_ approach at it makes simpler to iterate to all the lines (`part.split("/r/n")`) and deal with the two groups in a different way

samples/junit-tests/src/test/java/com/yelp/codegen/generatecodesamples/PostFileEndpointTest.kt
macisamuele: See previous comment.

`content` should contain the content of the uploaded file, so we need to make sure that we're continuing iterating on `part` from the correct position

Dropping this as the discussion on the point above will provide valuable input for this as well (modifying headers extraction implies modifying body extraction)

samples/junit-tests/src/test/java/com/yelp/codegen/generatecodesamples/PostFileEndpointTest.kt
macisamuele: I was thinking to use sequences in order to avoid the creation of the `String` related to the second part of the split as we don't use it and it is basically all the body.
I can agree that we're writing a test, that is going to ingest a limited amount of data, so it should be ok ™️

samples/junit-tests/src/test/java/com/yelp/codegen/generatecodesamples/PostFileEndpointTest.kt
cortinico: > I was using the iterator to traverse the all input only one.

I mean, it's inside a test 😅 readability >> perfomance.
I'd rather go for the functional approach rather than using an iterator over here.

cortinico: Also this could work for the content:

```
val content  = headers.entries.fold("") { previous, current -> "$previous\r\n${current.key}: ${current.value}" }
```

macisamuele: I see your point, I'm going to use a more functional style to implement this function.

I think that we have some misunderstanding here, so I'll write down an example to be more explicative.

part will look like
```
--BOUNDARY\r\n
header1: value1\r\n
header2: value2\r\n
\r\n
fileContent\r\n

```
and we want to extract:
 * `mapOf("header1" to "value1", "header2" to "value2")`  into `headers`
 * `fileContent` into `content`

PR #102, opened 2020-02-07
Add Gradle Lazy Configuration take n° 2

### LINKED ISSUE #63: Update the Gradle Plugin API to use Lazy Configuration

martinbonnin:
Gradle lazy configuration based on https://github.com/Yelp/swagger-gradle-codegen/pull/69.

Fixes #69 
Fixes #63


gradle-plugin/plugin/src/main/java/com/yelp/codegen/plugin/FeatureConfiguration.kt
```diff
@@ -1,11 +1,11 @@
...
+import org.gradle.api.Project
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.ListProperty
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.Optional
 
-open class FeatureConfiguration {
-
-    @Input
-    @Optional
```
martinbonnin: I'm not sure why this was needed. I don't think something else than task properties can be annotated.

gradle-plugin/plugin/src/main/java/com/yelp/codegen/plugin/GenerateTask.kt
```diff
@@ -20,59 +26,65 @@ const val DEFAULT_NAME = "defaultname"
 const val DEFAULT_PACKAGE = "com.codegen.default"
 const val DEFAULT_OUTPUT_DIR = "/gen"
 
-open class GenerateTask : DefaultTask() {
+abstract class GenerateTask : DefaultTask() {
```
martinbonnin: Making this class abstract allows to leverage Gradle service injection to create Properties automagically: https://docs.gradle.org/current/userguide/custom_gradle_types.html#service_injection

gradle-plugin/plugin/src/main/java/com/yelp/codegen/plugin/GenerateTask.kt
```diff
@@ -83,13 +95,13 @@ open class GenerateTask : DefaultTask() {
             specName ${'\t'} ${specName ?: "[ DEFAULT ] $DEFAULT_NAME"}
             specVers ${'\t'} ${specVersion ?: "[ DEFAULT ] $DEFAULT_VERSION"}
             input ${"\t\t"} $inputFile
-            output ${"\t\t"} ${outputDir ?: "[ DEFAULT ] $defaultOutputDir"}
+            output ${"\t\t"} ${outputDirectory ?: "[ DEFAULT ] $defaultOutputDir"}
             groupId ${'\t'} ${packageName ?: "[ DEFAULT ] default"}
             artifactId ${'\t'} ${packageName ?: "[ DEFAULT ] com.codegen"}
-            features ${'\t'} ${features?.headersToRemove?.joinToString(", ")?.ifEmpty { "[  EMPTY  ]" }}
+            features ${'\t'} ${headersToRemove.joinToString(separator = ",", prefix = "[", postfix = "]")}
```
martinbonnin: There's a small behaviour change here: if headersToRemove is empty, it will show `[]` instead of `[  EMPTY  ]`

samples/kotlin-android-moshi-codegen/build.gradle.kts
```diff
@@ -44,14 +44,14 @@ dependencies {
...
+    platform.set("kotlin")
+    packageName.set("com.yelp.codegen.samples")
+    specName.set("sample_specs")
+    specVersion.set("1.0.0")
+    inputFile.set(file("../sample_specs.json"))
+    outputDir.set(file("./src/main/java/"))
     features {
-        headersToRemove = arrayOf("Accept-Language")
+        headersToRemove.add("Accept-Language")
     }
```
martinbonnin: Moving to lazy configuration changes the plugin API. This certainly requires a major version, some warnings or something else.

gradle-plugin/plugin/src/main/java/com/yelp/codegen/plugin/FeatureConfiguration.kt
martinbonnin: It turns out this was into a @Nested task property so annotation are needed. IntelliJ complained because it required `@gete:Input` and not just `Input`

gradle-plugin/plugin/src/main/java/com/yelp/codegen/plugin/GenerateTask.kt
```diff
@@ -83,13 +92,13 @@ open class GenerateTask : DefaultTask() {
             specName ${'\t'} ${specName ?: "[ DEFAULT ] $DEFAULT_NAME"}
             specVers ${'\t'} ${specVersion ?: "[ DEFAULT ] $DEFAULT_VERSION"}
```
cortinico: `?: "[ DEFAULT ] $DEFAULT_VERSION` should go

cortinico: Same as line `109` (that I can't comment):
`params.add(specVersion ?: DEFAULT_VERSION)`

martinbonnin: Not sure I follow, you want to remove all the leading "[ DEFAULT ]" from the logs ? That's fine with me but I think we should do it for all params ?

gradle-plugin/plugin/src/main/java/com/yelp/codegen/plugin/GenerateTask.kt
```diff
@@ -83,13 +92,13 @@ open class GenerateTask : DefaultTask() {
...
-            output ${"\t\t"} ${outputDir ?: "[ DEFAULT ] $defaultOutputDir"}
+            output ${"\t\t"} ${outputDirectory ?: "[ DEFAULT ] $defaultOutputDir"}
             groupId ${'\t'} ${packageName ?: "[ DEFAULT ] default"}
             artifactId ${'\t'} ${packageName ?: "[ DEFAULT ] com.codegen"}
-            features ${'\t'} ${features?.headersToRemove?.joinToString(", ")?.ifEmpty { "[  EMPTY  ]" }}
+            features ${'\t'} ${headersToRemove.joinToString(separator = ",", prefix = "[", postfix = "]")}
         """.trimIndent())
 
-        val packageName = packageName ?: DEFAULT_PACKAGE
+        packageName = this.packageName.orNull ?: DEFAULT_PACKAGE
```
trevjonez: another cool part of the property api's in gradle is the ability to have a `convention` on them.

helps keep things more declarative and removes the need to check for null and such later on.
https://docs.gradle.org/current/javadoc/org/gradle/api/provider/Property.html#convention-T-

there is both a `convention(T)` and `convention(Provider<T>)` so if it is something that needs to be lazily calculated you can still do that at a higher level and keep this task execution code more streamlined.

cortinico: That looks great! 

@martinbonnin would you be able to integrate this suggestion?

gradle-plugin/plugin/src/main/java/com/yelp/codegen/plugin/GenerateTask.kt
cortinico: Sorry I wasn't clear, my fault.
The `specVersion` variable is always not null. So the elvis operator is not needed and the "null" branch will never run (so you can remove it).

samples/kotlin-android-moshi-codegen/build.gradle.kts
cortinico: Yup I've added this PR to the `2.0.0` milestone

samples/kotlin-android/build.gradle.kts
```diff
@@ -37,14 +37,14 @@ dependencies {
...
-    specName = "sample_specs"
-    specVersion = "1.0.0"
-    inputFile = file("../sample_specs.json")
-    outputDir = file("./src/main/java/")
+    platform.set("kotlin")
+    packageName.set("com.yelp.codegen.samples")
+    specName.set("sample_specs")
+    specVersion.set("1.0.0")
+    inputFile.set(file("../sample_specs.json"))
+    outputDir.set(project.layout.buildDirectory.dir("./src/main/java/"))
```
cortinico: If you add the build directory, then please place the files inside `/gen` or `/generated`.

gradle-plugin/plugin/src/main/java/com/yelp/codegen/plugin/GenerateTask.kt
martinbonnin: Indeed :+1: . IntelliJ did it for me after I pressed Alt + Enter :)

samples/kotlin-android/build.gradle.kts
martinbonnin: I reverted it back to `projectDirectory.dir(".src/main/java"))`. 
It feels a bit weird to have generated files under version control though. Maybe we should move eveything under `/build` ?

gradle-plugin/plugin/src/main/java/com/yelp/codegen/plugin/GenerateTask.kt
martinbonnin: Looking at the `convention`, thing, I simplified a bit the logs code but a side effect is that the '[ DEFAULT ]' indication is lost for all properties. Let me know if that's ok or if I should put it back.

gradle-plugin/plugin/src/main/java/com/yelp/codegen/plugin/GenerateTask.kt
martinbonnin: Not sure how conventions would help here as we overwrite the task properties from the `GradleTaskConfiguration` object and this will remove any previously set convention.

@trevjonez did you have something specific in mind ?

trevjonez: In that case you could set the defaults as the convention on the property of the gradle task configuration object. 

Looking at this more, I think most if not all of the properties on the task should not be marked optional since a value is required for the task to run. My thinking is that not having the defaults in the property when the task is checked before running it might open you up to odd build cache key issues.

samples/kotlin-android/build.gradle.kts
trevjonez: +1 to not placing generated files in VC. The whole point of swagger is to be schema driven, if you check in generated outputs, you run the risk of being out of sync with the schema, which sort of defeats the key purpose of this.

In this case if the inputs and outputs of the gradle tasks are setup correctly, the gradle build cache will avoid the work in a similar way that checking things in would, but without the risk of having your inputs and outputs getting out of sync.

gradle-plugin/plugin/src/main/java/com/yelp/codegen/plugin/GenerateTask.kt
martinbonnin: Yea, good point. That's a bigger change but I guess now is the time :). I'll look into it.

gradle-plugin/plugin/src/main/java/com/yelp/codegen/plugin/GenerateTask.kt
```diff
@@ -20,112 +23,111 @@ const val DEFAULT_NAME = "defaultname"
...
-    @Optional
-    @Option(option = "outputDir", description = "Configures path of the Generated code directory.")
-    var outputDir: File? = null
-
-    @InputFiles
-    @Optional
-    @Option(option = "extraFiles",
+    @get:Input
+    @get:Optional
+    @get:Option(option = "platform", description = "Configures the platform that is used for generating the code.")
```
trevjonez: another way to put these could be:
```kotlin
@get:[Input Option(...)]
```
https://kotlinlang.org/docs/reference/annotations.html#annotation-use-site-targets
more of a style choice than anything, just throwing it out there.

samples/kotlin-android/build.gradle.kts
cortinico: Agree, that's also the reason why `/src/main/java` is inside the `.gitignore`. I'm fine either way as long as we don't commit them.

gradle-plugin/plugin/src/main/java/com/yelp/codegen/plugin/GenerateTask.kt
martinbonnin: Changes done, let me know what you think.

samples/kotlin-android/build.gradle.kts
martinbonnin: Some of the `junit-test` generated code is commited. See https://github.com/Yelp/swagger-gradle-codegen/tree/master/samples/junit-tests/src/main/java

Is that intentional ?

gradle-plugin/plugin/src/main/java/com/yelp/codegen/plugin/GenerateTask.kt
martinbonnin: I'm new to this `Option` annotation actually. Is that a common thing to do ? Define task inputs from the command line ?

cortinico: > I'm new to this Option annotation actually

In our case we wanted to have this setup inside a docker container. So having a command line API made easier to configure the task from an external environment.

gradle-plugin/plugin/src/main/java/com/yelp/codegen/plugin/GenerateTaskConfiguration.kt
```diff
@@ -1,18 +1,44 @@
...
+    val outputDir = objects.directoryProperty().convention(project.layout.buildDirectory.dir("gen").get())
+
+    val extraFiles = objects.directoryProperty()
+    val features: FeatureConfiguration = FeatureConfiguration(objects)
 
     fun features(action: Action<FeatureConfiguration>) = action.execute(features)
+
+    companion object {
+        private fun readVersionFromSpecfile(specFile: File): String {
+            val swaggerSpec = SwaggerParser().readWithInfo(specFile.absolutePath, listOf(), false).swagger
```
cortinico: Still not 100% sure that this method belongs here. This currently pulls in an `import io.swagger.parser.SwaggerParser` that I would love to avoid having in the `GenerateTaskConfiguration` file. How about having it as a top level function in a separate file (à la Util file)?

martinbonnin: I moved that to `VersionHelper`, couldn't think of a better name but happy to change it.

Doing this actually got me wondering why the version was needed as input in the first place if it's already present in the spec file ? What are the use cases to override that ? Can a specFile contain several versions of an api ?

cortinico: > why the version was needed as input in the first place if it's already present in the spec file 

Ideally you should not override the version in the gradle file and use the one provided in the spec file. However there are 2 scenarios that made us think that this could have been a case:

- Updating the generator code/template. Say that you bump the `swagger-gradle-codegen` version and want to re-publish an artifact generated from the updated generator with a different scenario.

- Semantic versioning. We thought about having the final version as a composition of the version of the generator + the version of the spec. So you could have something like `<generator-version>.<spec-version>.<patch>`. Moreover the spec version could be a timestamp (say `2020010101`) and you would probably want to have a way to beautify it.

Generally the idea of having a way to override the version felt useful.

cortinico:
Thanks for the great change @martinbonnin 🚀

PR #105, opened 2020-02-08
Migrate Makefile targets to Gradle tasks

martinbonnin:
Replace the Makefile targets with Gradle tasks:

```
make install-hooks => ./gradlew installHooks
make run-hooks => ./gradlew runHooks
make test => ./gradlew preMerge
```


build.gradle.kts
```diff
@@ -5,3 +5,61 @@ subprojects {
...
+}
+
+val preMerge = tasks.register("preMerge") {
+    description = "Runs pre-commit hooks, build the plugin and sample and execute tests."
+
+    dependsOn(runHooks)
+    dependsOn(gradle.includedBuild("gradle-plugin").task(":plugin:build"))
+    dependsOn(gradle.includedBuild("gradle-plugin").task(":plugin:check"))
+    dependsOn(subprojects.filter { it.name != "samples" }.map { it.tasks.named("assembleDebug") })
+    dependsOn(subprojects.filter { it.name != "samples" }.map { it.tasks.named("check") })
```
cortinico: This won't work at the moment, as you have to specify also the `mustRunAfter` rules. In the current situation you're just saying that `preMerge` depends on `runHooks`, `:plugin:build`, `:plugin:check`, `assembleDebug`, `check` but you're not specifying the order.

This is also the reason why Travis is failing (he's running a one of a sample's `check` before the `generateSwagger`).

build.gradle.kts
```diff
@@ -5,3 +5,61 @@ subprojects {
...
+}
+
+subprojects {
+    afterEvaluate { // Remove when we have lazy configuration
+        tasks.all {
+            if (name == "assembleDebug") {
+                // assembleDebug needs the generated files
+                dependsOn(tasks.named("generateSwagger"))
+            }
```
cortinico: Maybe we can have this as part of the `preMerge`?

build.gradle.kts
martinbonnin: > This is also the reason why Travis is failing (he's running a one of a sample's check before the generateSwagger).

That was the point of this block int the root build.gradle.kts: 

```kotlin
        tasks.all {
            if (name == "assembleDebug") {
                dependsOn(tasks.named("generateSwagger"))
            }
        }
```

Except I forgot that the generated files are needed before the `assemble` task. I hopefully fixed that here: https://github.com/Yelp/swagger-gradle-codegen/pull/105/commits/b0e45e7a011b7ed44d9758907cd1888fbcd3f630#diff-392475fdf2bc320d17762ed97109a121R66

build.gradle.kts
martinbonnin: I guess we could but isn't it nice to have gradle regenerate the swagger files on-demand ? 

I.e. you could run `assembleDebug`, `build`,  `compileMainReleaseeKotlin`, etc... and Gradle would make sure the generated files are up-to-date without having to run `generateSwagger` first.

build.gradle.kts
```diff
@@ -5,3 +5,68 @@ subprojects {
...
+    commandLine("./venv/bin/pre-commit", "run", "--all-files")
+}
+
+val preMerge = tasks.register("preMerge") {
+    description = "Runs pre-commit hooks, build the plugin and sample and execute tests."
+
+    dependsOn(runHooks)
+    dependsOn(gradle.includedBuild("gradle-plugin").task(":plugin:build"))
+    dependsOn(gradle.includedBuild("gradle-plugin").task(":plugin:check"))
+    dependsOn(subprojects.filter { it.name != "samples" }.map { it.tasks.named("assembleDebug") })
```
martinbonnin: We could maybe remove the "assembleDebug" there. The "check" test below already executes `testDebugUnitTest` and `testReleaseUnitTest` so it will compile the sources for both debug and release mode.

macisamuele:
@martinbonnin ++
Thanks for putting the effort to gradle-ify the repo 😄 

I'm not very familiar with gradle so I'll rely @cortinico for comments.
For what I can see the makefile targets have been converted to gradle tasks , so all good here

build.gradle.kts
cortinico: Agree, let's remove it 👍

Issue #109, opened 2020-02-11
Handle null values in Kotlin TypesAdapter


mikeyshean:
Example:  Null values for BigDecimal in request body will throw "Nesting Problem" error.   Description of root cause is similar to the issue described here: https://github.com/square/moshi/issues/806

The TypesAdapter will need explicit handling for null values to prevent this.

macisamuele:
Asking here, as the issue looks a better place to talk about it.

Could you please publish a minimal spec example that triggers the issue?

cortinico:
> Could you please publish a minimal spec example that triggers the issue?

A failing JUnit test would also suffice

djensen47:
It might take us a while to construct a useful JUnit test because the YAML triggering the error is just too big to use as a test case: https://developer.sabre.com/sites/default/files/eWejJIu8.yaml

(The attribute that triggered the issue on serialization was `OnTimeRate`)

So we'll have to construct something artificial but similar.

macisamuele:
> So we'll have to construct something artificial but similar.

That's definitely OK. Having an additional endpoint into [junit_tessts_specs.json](https://github.com/Yelp/swagger-gradle-codegen/blob/master/samples/junit-tests/junit_tests_specs.json) would be great.

I might be supportive on identifying the minimal example. Something that would help is having a small example on how you were interacting with the API that lead to the identification of the issue.
cc: @djensen47

macisamuele:
ping @djensen47 @mikeyshean

djensen47:
Sorry, your ping got lost in a river of Github notifications. 

We're not going to have cycles to create a proper reproduction of the bug until early or mid-April. Same goes for updating the proposed PR. We have a workaround in place but since it's a bit of a hack we definitely want to get this fixed properly in the library. I'm actually creating a ticket in our Jira system to revisit and help fix this.

cortinico:
> We're not going to have cycles to create a proper reproduction of the bug until early or mid-April. Same goes for updating the proposed PR. We have a workaround in place but since it's a bit of a hack we definitely want to get this fixed properly in the library. I'm actually creating a ticket in our Jira system to revisit and help fix this.

Awesome, keep us posted 👍

stew-wwt:
Any update this issue? We ran into the same issue implementing the plugin in our project today

cortinico:
> Any update this issue? We ran into the same issue implementing the plugin in our project today

[...]

PR #110, opened 2020-02-11
Add null value handling for kotlin TypesAdapters

mikeyshean:
PR for #109

### DIFF
--- a/gradle-plugin/plugin/src/main/resources/kotlin/tools/TypesAdapters.kt.mustache
+++ b/gradle-plugin/plugin/src/main/resources/kotlin/tools/TypesAdapters.kt.mustache
@@ -60,7 +60,7 @@ internal class LocalDateAdapter : XNullableJsonAdapter<LocalDate>() {
     override fun fromNonNullString(nextString: String): LocalDate = LocalDate.parse(nextString, formatter)
 
     override fun toJson(writer: JsonWriter, value: LocalDate?) {
-        value?.let { writer.value(it.format(formatter)) }
+        value?.let { writer.value(it.format(formatter)) } ?: writer.nullValue()
     }
 }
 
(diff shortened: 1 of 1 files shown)


djensen47:
@cortinico Any thoughts on this PR?

We spent a day and a half debugging this issue and @mikeyshean discovered that `toJson` needs to handle null values otherwise it triggers an issue in Moshi thinking there is a nesting problem.

macisamuele:
The pr looks sane to me, but I need some time to verify it.
It looks surprising that we haven't face this issue before.

Something that for sure I'll ask is to have testing coverage over the change and ideally a unit test in samples/junit-test that ensures no further regressions around this.

gradle-plugin/plugin/src/main/resources/kotlin/tools/TypesAdapters.kt.mustache
```diff
@@ -60,7 +60,7 @@ internal class LocalDateAdapter : XNullableJsonAdapter<LocalDate>() {
     override fun fromNonNullString(nextString: String): LocalDate = LocalDate.parse(nextString, formatter)
 
     override fun toJson(writer: JsonWriter, value: LocalDate?) {
-        value?.let { writer.value(it.format(formatter)) }
+        value?.let { writer.value(it.format(formatter)) } ?: writer.nullValue()
```
macisamuele: I would rather suggest to have concrete implementations dealing only with the "not-nullable" case and having `XNullableJsonAdapter` taking care of the `null` handling.

Something like
```kotlin
internal abstract class XNullableJsonAdapter<T> : JsonAdapter<T>() { 
    abstract fun fromNonNullString(nextString: String): T
    abstract fun toJsonNonNull(writer: JsonWriter, value: T)
    ...
}
```
This allows to focus the attention of nullable in the class related to nullable and to the `T` type encoding in the related adapter class

djensen47:
We were able to trigger this with a swagger definition that had a deeply nested BigDecimal value. In the swagger specification the type was set as "number".

klaszlo8207:
Any news on this?

djensen47:
Sorry, the project we were using this on was canceled. @klaszlo8207 feel free to pick up this PR and add a regression test to it.

Issue #108, opened 2020-02-10
Unable to load class 'org.gradle.kotlin.dsl.precompile.v1.PrecompiledProjectScript'.


mariuspena:
Hi. 
made a new project, gradle version 3.5.3 (**<<< EDIT: actually it is 5.4.1**)
added in the the module gradle file:
```
buildscript {
    repositories {
        maven { url "https://plugins.gradle.org/m2/" }
    }
    dependencies {
        classpath "com.yelp.codegen:plugin:1.3.0"
    }
}
apply plugin: "com.yelp.codegen.plugin"
generateSwagger {
    platform = "kotlin"
    packageName = "com.yelp.codegen.samples"
    inputFile = file("./sample_specs.json")
    outputDir = file("./src/main/java/")
}
```

but I get this error:
Unable to load class 'org.gradle.kotlin.dsl.precompile.v1.PrecompiledProjectScript'.

can somebdoy help me ? thx!

cortinico:
Hi @mariuspena, thanks for using this project.

> made a new project, gradle version 3.5.3

As described in the README at the "Requirements" section, you need to have Gradle 5+ to use this plugin:
https://github.com/Yelp/swagger-gradle-codegen/blob/master/README.md#L55-L60

We can't support you further till you update Gradle.

mariuspena:
my bad. my gradle version is: 5.4.1

cortinico:
Thanks for the update

> added in the the module gradle file:

Can you please provide more context? Are you able to link your project?
Could you please post both the root project gradle file and the module gradle file?

My gut feeling is that this failure is unrelated to the plugin, but I don't have enough context to tell you at this point.

mariuspena:
top level gradle file:
`// Top-level build file where you can add configuration options common to all sub-projects/modules.

buildscript {
    ext.kotlin_version = '1.3.61'
    repositories {
        google()
        jcenter()
        maven { url "https://jitpack.io" }
    }
    dependencies {
        classpath 'com.android.tools.build:gradle:3.5.3'
        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
        // NOTE: Do not place your application dependencies here; they belong
        // in the individual module build.gradle files
    }
}

allprojects {
    repositories {
        google()
        jcenter()

    }
}

task clean(type: Delete) {
    delete rootProject.buildDir
}
`
and the module gradle file:
`apply plugin: 'com.android.application'
[...]

PR #111, opened 2020-02-12
Miscellaneous documentation update

macisamuele:
The goal of this PR is:
 * highlight Gradle5 incompatibility of release 1.3.0
 * re-establish Gradle5+ support
 * remove all traces of `make ...` invocations (as Makefile was removed in #105 )

NOTE: Ideally we should be able to merge this once #108 troubleshooting is over.
Ideally, we should not drop support to gradle versions if we have no benefits on doing it

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,7 +15,7 @@ Here you can find the release notes for this project. Please note that list of r
 * ⚙️ Gradle to 5.6.2 [#67]
 * ⚙️ Add JaCoCo for CodeCov [#66]
 
-⚠️: `swagger-gradle-codegen:plugin:1.3.0` is **NOT** supported on Gradle 5
+⚠️: `swagger-gradle-codegen:plugin:1.3.0` is supported **ONLY** on Gradle 6+
 
 Thanks to @cortinico @doug-precocity @redwarp @macisamuele @filipemp for the support with this release
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,8 @@ Here you can find the release notes for this project. Please note that list of r
 * ⚙️ Gradle to 5.6.2 [#67]
 * ⚙️ Add JaCoCo for CodeCov [#66]
 
+⚠️: `swagger-gradle-codegen:plugin:1.3.0` is **NOT** supported on Gradle 5
+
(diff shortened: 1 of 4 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.5-all.zip
```
cortinico: I'd revert to `5.6.2` possibly.
I'm also in favor for keeping Gradle 6.1 and just letting our users know that they need to update Gradle.

README.md
```diff
@@ -156,15 +156,15 @@ We also recommend you set up:
 Before starting developing, please run:
 
 ```
-make install-hooks
+./gradlew installHooks
 ```
 
 This will take care of installing the pre-commit hooks to keep a consistent style in the codebase.
 
 While developing, you can run tests on the plugin using:
```
cortinico:
```suggestion
While developing, you can build, run pre-commits, checks & tests on the plugin using:
```

gradle/wrapper/gradle-wrapper.properties
macisamuele: I'm going to preserve 5.5 as it was the version used in 1.2.0  (https://github.com/Yelp/swagger-gradle-codegen/blob/1.2.0/gradle/wrapper/gradle-wrapper.properties#L3)

CHANGES.md
```diff
@@ -15,6 +15,8 @@ Here you can find the release notes for this project. Please note that list of r
 * ⚙️ Gradle to 5.6.2 [#67]
 * ⚙️ Add JaCoCo for CodeCov [#66]
 
+⚠️: `swagger-gradle-codegen:plugin:1.3.0` is **NOT** supported on Gradle 5
```
cortinico: Maybe rewording to

```suggestion
⚠️: `swagger-gradle-codegen:plugin:1.3.0` is supported **ONLY** on Gradle 6+
```

PR #112, opened 2020-02-13
Remove the `kotlin-dsl` plugin

martinbonnin:
Closes https://github.com/Yelp/swagger-gradle-codegen/issues/108

It turns out the `kotlin-dsl` plugin embeds code that's using Gradle 6.1 classes while we shouldn't really need this. Remove it and use the plain `java-gradle-plugin` instead.

Also add some checks and tests about Gradle version.


README.md
```diff
@@ -56,7 +56,7 @@ Once you setup the plugin correctly, you can call the `:generateSwagger` gradle
 
 In order to run this gradle plugin you need to fulfill the following requirements:
 
-* Gradle 5.x - This plugin uses Gradle 5 features, and you will need to setup your Gradle wrapper to use 5.0 or more.
+* Gradle 5.x - This plugin uses Gradle 5 features, and you will need to setup your Gradle wrapper to use 5.4.1 or more.
```
martinbonnin: I bumped from `5.0` to `5.4.1` because the testProject fails on `5.0` due to:

```
Build file '/home/martin/git/swagger-gradle-codegen/gradle-plugin/plugin/junit7355662627637669365/project/build.gradle.kts' line: 9

* What went wrong:
Script compilation errors:

  Line 09:       content {
                 ^ Unresolved reference: content

  Line 10:         excludeGroup("com.yelp.codegen")
                   ^ Unresolved reference: excludeGroup
```

I think it should be safe to assume most people are using gradle >= 5.4.1 but we can certainly remove that limitation if really needed.

cortinico:
> @martinbonnin I would personally prefer to use kotlin-dsl as far as possile.

+1

martinbonnin:
@macisamuele no problem if you want to keep it but as you said, it's bringing a downside and I don't see the upside. The amount of code is almost similar and it might bring things like `HasImplicitReceiver` to allow to use `this` instead of `it` but this almost sounds like a downside too. For someone used to Kotlin, this transformation looks a bit magic.
Or are there more advantages ?

cortinico:
> Or are there more advantages ?

`kotlin-dsl` plugin allows us to use the precompiled plugins:
https://docs.gradle.org/current/userguide/kotlin_dsl.html#kotdsl:precompiled_plugins
(They actually added a note that covers exactly our scenario)

That is convenient as it simplifies our Gradle setup. I'll try to stick to it as long as it's possible. Although, I don't have a really strong opinion on this, we could even drop if you have a strong argument @martinbonnin 👍

gradle-plugin/plugin/src/test/java/com/yelp/plugin/PluginTests.kt
```diff
@@ -32,4 +32,20 @@ class PluginTests {
                 .build()
         Assert.assertEquals(TaskOutcome.UP_TO_DATE, result2ndRun.task(":generateSwagger")?.outcome)
     }
+
+    @Test
+    fun testGradle5() {
```
macisamuele: nit. I would rename this test to something like "testMinSupportdGradleVersion" or something similar, just to make sure that we'll not be surprised in the future about what is this test doing

cortinico:
Following up the discussions we had on Slack [here](https://kotlinlang.slack.com/archives/CU233PG2Z/p1581792288012400)

We decided for replacing `kotlin-dsl` plugin with `java-gradle-plugin` for the reasons aforementioned by @martinbonnin.
I'm merging this and will prepare an upcoming release soon 👍

PR #114, opened 2020-02-15
Add plugin tests against an older version of Gradle

martinbonnin:
This is taken from https://github.com/Yelp/swagger-gradle-codegen/pull/112 but still keeping the `kotlin-dsl` plugin.


cortinico:
Discarding as #112 was merged

PR #118, opened 2020-02-17
Prepare Release v1.4.0

cortinico:
We do have a lot of stuff on master (e.g. Moshi codegen). Let's create a new release 👍

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -54,7 +54,7 @@ val preMerge = tasks.register("preMerge") {
 
 plugins {
     id("com.android.library").version("3.5.3").apply(false)
-    id("com.yelp.codegen.plugin").version("1.3.0").apply(false)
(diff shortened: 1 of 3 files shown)


build.gradle.kts
```diff
@@ -54,7 +54,7 @@ val preMerge = tasks.register("preMerge") {
 
 plugins {
     id("com.android.library").version("3.5.3").apply(false)
-    id("com.yelp.codegen.plugin").version("1.3.0").apply(false)
+    id("com.yelp.codegen.plugin").version("1.4.0").apply(false)
```
macisamuele: Would be possible to reference `PLUGIN_VERSION` here instead of hardcoding the value?

This is not a blocking request but it would be nice if releasing a new version requires one less step :)

cortinico: Looks like we can't as `PLUGIN_VERSION` is inside the `buildSrc` of the `plugin` build and is not accessible from the outer build.

PR #127, opened 2020-03-05
Drop support for moshi-kotlin (reflective approach for JSON encoding/decoding)

### LINKED ISSUE #83: Remove support for moshi-reflective approach
PR #82 has added support for code-generated Moshi Adapters.
As result of the [discussion on the PR](https://github.com/Yelp/swagger-gradle-codegen/pull/82#discussion_r369482258) we need clear up the current mixed support of reflective and moshi-codegen approach.

We need to do it in order to:
 * reduce complexity
 * reduce dependencies (as mentioned in #62 `moshi-kotlin` depends on `kotlin-reflect` which is pretty big)
 * provide better performances with the least surprises for user

In order to achieve such objective we need to:
- [ ] Remove `kotlin-android-with-moshi-kotlin-codegen` and update `kotlin-android` to use only `moshi-kotlin-codegen`
- [ ] Update `groovy-android`
- [ ] Improve documentation to highlight a possible approach to still use reflective approach (a way to rebuild `GeneratedCodeConverters.moshi`)

**Disclaimer**: The list is not supposed to be complete, so if I'm missing something please update the checklist ;)

macisamuele:
This PR is aimed to fix #83 

In order to be ready for the next major release we need (according to the milestones) to remove the support of moshi-kotlin (reflection) and ensure that moshi-codegen is the default appraoch.

Still preserving the feature does not require a lot of effort and still there might be use-cases were it would be preferred to use reflection.

So the main idea here is to enable moshi-kotlin via a feature (as we do support them) :)


README.md
```diff
@@ -138,7 +135,8 @@ Here a list of all the supported features:
 
 | Feature | Description | Command line |
 | -------- | ----------- | ------------ |
-| `headersToRemove` | List of headers that needs to be ignored for every endpoints. The headers in this list will be dropped and not generated as parameters for the endpoints. | `--featureHeaderToRemove=` |
+| `headersToRemove` | List of headers that needs to be ignored for every endpoints. The headers in this list will be dropped and not generated as parameters for the endpoints. | `-ignoreheaders=` |
+| `supportMoshiReflective` | Boolean value that enables the usage of [`moshi-kotlin`](https://github.com/square/moshi#reflection) into the generated code.<br />By default it is not enabled. | `-supportMoshiReflective` |
```
cortinico: As discussed offline, I'd rather won't have this feature right now. Let's keep the template `moshi-codegen` only. We can still add it if we get a feature request for it.

macisamuele: I agree ... I'll keep the diff on a gist such that if we will need to re-add it it won't take very long.

I'll update the branch removing the new feature and references to it

PR #129, opened 2020-05-15
Detekt to 1.9.0

cortinico:
Just keeping our Detekt version up to date


macisamuele:
@cortinico update https://github.com/Yelp/swagger-gradle-codegen/blob/master/.github/workflows/pre-merge.yaml#L16 to have green tests and then feel free to merge.

tip: as the python venv is needed to install pre-commit only, I would install `3.7` (without specifying the patch version).

PS. a list of actions have a new version, we might want to update that as well (actions/checkout, actions/setup-python

PR #131, opened 2020-05-15
Small GH Actions optimizations

cortinico:
- Set `fail-fast` to false so we'll see the outcome of all the builds
- Trigger `push` only on `master` so we won't have duplicate builds

### DIFF
--- a/.github/workflows/pre-merge.yaml
+++ b/.github/workflows/pre-merge.yaml
@@ -1,9 +1,16 @@
 name: Pre Merge Checks
-on: [push, pull_request]
+on: 
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - '*'
 
 jobs:
   gradle:
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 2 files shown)


.github/workflows/pre-merge.yaml
```diff
@@ -1,9 +1,16 @@
 name: Pre Merge Checks
-on: [push, pull_request]
+on: 
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - '*'
```
macisamuele, line 8: nit. If I remember correctly you might just ignore the `- '*'` line ;)

cortinico:
> Side note:
> I generally discourage people contributing to projects to create branches on the project itself.
> I tend to suggest to use the local fork to perform work and then open PRs to merge new features.

Agree, but this is not possible if you have write access to the repo + you're editing the file from the browser (like I do for small changes like this one).

.github/workflows/pre-merge.yaml
cortinico: :( 
Screenshot 2020-05-15 at 18 24 38

PR #134, opened 2020-06-06
Allow tests execution on environments there Android SDK is not installed

macisamuele:
The goal of this PR, as mentioned in the title, is to detach the project from Android.

The Android dependencies are mostly needed for the sample projects, which gives an idea on how to use/integrate the library into end-user scenarios, but is not needed to test the overall library functionality.

In order to achieve this objective the inclusion of all the samples is dependent on the availability of an Android SDK (via env variable or via system property).
Anyway we want to guarantee that Android SDK is installed while running tests on CI, and so we do consider Android SDK present if on CI (almost all the public CI systems do expose `CI=true` environmental variable).

@cortinico: This is published as draft to have your feedback on the idea and eventually have your input on how to check the presence of `sdk.dir` from `local.properties` file (I'd like doing it via gradle rather than `cat ... | grep ...` ;) )


settings.gradle.kts
```diff
@@ -17,10 +17,24 @@ pluginManagement {
...
 includeBuild("gradle-plugin")
+
+fun isAndroidSDKConfigured(): Boolean {
+    if (System.getenv("ANDROID_SDK_ROOT") != null) {
+        return true;
+    }
+    // TODO: Check if `sdk.dir` property is defined in `local.properties`
+
+    // Ensure that on CI systems we do assume that Android is installed
+    return System.getenv("CI") != null);
```
cortinico, line 39:
```suggestion
    return System.getenv("CI") != null;
```

settings.gradle.kts
```diff
@@ -17,10 +17,24 @@ pluginManagement {
...
         ":samples:kotlin-coroutines",
-        ":samples:groovy-android")
+        ":samples:groovy-android"
+    )
+}
 
 includeBuild("gradle-plugin")
+
+fun isAndroidSDKConfigured(): Boolean {
+    if (System.getenv("ANDROID_SDK_ROOT") != null) {
```
cortinico, line 33:
```suggestion
    if (System.getenv("ANDROID_SDK_ROOT") != null || System.getenv("ANDROID_HOME") != null) {
```

settings.gradle.kts
```diff
@@ -17,10 +17,24 @@ pluginManagement {
...
+        return true;
+    }
+    // TODO: Check if `sdk.dir` property is defined in `local.properties`
```
cortinico, line 36: You can remove this. If `sdk.dir` property is not defined in `local.properties`, it's up to the user to configure it. The ID or the command line will notify if the environment is misconfigured.

macisamuele: The idea of this PR is to conditionally include the projects depending on the presence of the android SDK.
The projects should be included if the Android SDK is available (`ANDROID_SDK_ROOT` env variable or `sdk.dir` property) or it's running on the CI 

Considering this we need read the property to ensure that it is present.

settings.gradle.kts
macisamuele: I'm surprised about `ANDROID_HOME`. When running it (on a machine without android) I get an error message referring to `ANDROID_SDK_ROOT` but with NO mentions to `ANDROID_HOME`

cortinico: From: https://developer.android.com/studio/command-line/variables

```
ANDROID_SDK_ROOT | Sets the path to the SDK installation directory. Once set, the value does not typically change, and can be shared by multiple users on the same machine. ANDROID_HOME, which also points to the SDK installation directory, is deprecated.
```

So yeah let's take both just for backward compat.
Also the GH runner has both set correctly.

settings.gradle.kts
cortinico: I can't find official docs right now, anyway the `local.properties` file will be recreated if missing with content of the `ANDROID_SDK_ROOT`. So it's sufficient to check for the env variables.

macisamuele: I've updated the PR reversing the logic (decide, via env variable, if android should be skipped) because there is no way to access to the content of `local.properties` within `settings.gradle.kts`.

This is relevant to me as I might define the Android SDK into the local properties file (and no env variable) and it would be _annoying_ if Android tasks would be ignored.

About _no way to access to the content_, technically there is a hack that can be used (`cat ... | grep sdk.dir`) but it look bad enough to read a file hoping that a string is present.

So I've decided to get a different direction here.

macisamuele:
The change is described in https://github.com/Yelp/swagger-gradle-codegen/pull/134#discussion_r437989354

PR #136, opened 2020-06-10
Use kotlinx.serialization instead of moshi

friedger:
This PR tries to make the project better suitable for multiplatform projects.

Therefore, moshi is replaced by kotlinx.serialization. Probably, this should go in a different folder or classes.

This is mainly to see what changes need to be done for kotlinx.serialization


cortinico:
Thanks for the PR @friedger, I think it's a good starting point to have a prototype (though I can't be merged as it's going to be a breaking change).

I think the goal could be to have the `junit-test` module to run correctly on CI. 

So far you have ktlint failing. I suggest you to install the pre-commit hooks with `./gradlew installHooks` so that you have the code formatted before pushing.

PR #138, opened 2020-07-27
Replace ThreeTenBp with Gradle core library desugaring

agologan:
This a proposal to replace the ThreeTenABP with the new desugared version especially since that's also the recommendation of the lib maintainer.

_Created in hopes of resolving a parser issue with the zoneId but the problem is the same in the desugared version._


cortinico:
Thanks for your contribution @agologan 

> Created in hopes of resolving a parser issue with the zoneId but the problem is the same in the desugared version.

Would you be able to link the issue, just to have more context and future reference?

README.md
```diff
@@ -56,7 +56,7 @@ Once you setup the plugin correctly, you can call the `:generateSwagger` gradle
 
 In order to run this gradle plugin you need to fulfill the following requirements:
 
-* Gradle 5.x - This plugin uses Gradle 5 features, and you will need to setup your Gradle wrapper to use 5.4.1 or more.
+* Gradle 6.x - This plugin uses Gradle 6 features, and you will need to setup your Gradle wrapper to use 6.1.1 or newer.
```
cortinico, line 59: Please amend this change. Though that is true for the source code on `master`, the latest released version (`v1.4.1`) works fine on Gradle 5.4.1.

If we merge this change, users will be confused by this discrepancy between the README and the latest stable. We can update it once we release `v2.x`

build.gradle.kts
```diff
@@ -55,7 +55,7 @@ val preMerge = tasks.register("preMerge") {
 }
 
 plugins {
-    id("com.android.library").version("3.5.3").apply(false)
+    id("com.android.library").version("4.0.0").apply(false)
```
cortinico, line 58:
```suggestion
    id("com.android.library").version("4.0.1").apply(false)
```

agologan:
> > Created in hopes of resolving a parser issue with the zoneId but the problem is the same in the desugared version.
> 
> Would you be able to link the issue, just to have more context and future reference?

Sure, it's about this one: https://stackoverflow.com/questions/43360852/cannot-parse-string-in-iso-8601-format-lacking-colon-in-offset-to-java-8-date which I'll fix by building my own version with a corrected date formatter.

cortinico:
> Sure, it's about this one: [stackoverflow.com/questions/43360852/cannot-parse-string-in-iso-8601-format-lacking-colon-in-offset-to-java-8-date](https://stackoverflow.com/questions/43360852/cannot-parse-string-in-iso-8601-format-lacking-colon-in-offset-to-java-8-date) which I'll fix by building my own version with a corrected date formatter.

Thanks for sharing 👍 

The PR looks good to me, though there are several things that needs to be addressed:
- This line can be removed https://github.com/Yelp/swagger-gradle-codegen/blob/e5b483556f01571c46cb840a77635de0b8efba02/samples/junit-tests/build.gradle#L21-L22
- This test should be updated with a different example (one from `java.time` or whatever other package works): https://github.com/Yelp/swagger-gradle-codegen/blob/e5b483556f01571c46cb840a77635de0b8efba02/gradle-plugin/plugin/src/test/java/com/yelp/codegen/KotlinGeneratorTest.kt#L245-L249

Other than that, we can merge it

PR #140, opened 2020-10-23
Update several dependencies

cortinico:
Keeping the project up to date:
- Gradle to 6.7
- Kotlin to 1.3.72
- Detekt to 1.14.2
- gradle.plugin-publish to 0.12.0
- Moshi to 1.11.0
- OkHttp to 3.12.12
- Retrofit to 2.9.0
- RxJava to 2.2.20
- JUnit to 4.13.1
- Mockwebserver 3.12.10


README.md
```diff
@@ -56,7 +56,7 @@ Once you setup the plugin correctly, you can call the `:generateSwagger` gradle
 
 In order to run this gradle plugin you need to fulfill the following requirements:
 
-* Gradle 5.x - This plugin uses Gradle 5 features, and you will need to setup your Gradle wrapper to use 5.4.1 or more.
+* Gradle 6.7+
```
macisamuele, line 59: I'm not sure about the fact that to run the plugin you need Gradle6.7+

We have a test that ensures the min dependency and seems it is working properly
https://github.com/Yelp/swagger-gradle-codegen/blob/master/gradle-plugin/plugin/src/test/java/com/yelp/plugin/PluginTests.kt#L43

Considering this I would suggest to avoid changing this line or updating the test to reflect the new min version to use.

macisamuele (commented):
Changes looks fine to me.

Just a few notes:
 * Please rebase your branch once #141  is merged
 * address nit of min gradle version to run the plugin

README.md
cortinico: You're correct. If the Gradle runner tests are passing, it's safe to assume is 5.4.1+ compatible.

macisamuele (approved):

PR #141, opened 2020-10-23
Update pre-commit hook versions

cortinico:
Bumping the pre-commit hook versions:
- pre-commit-hooks to 3.3.0
- detect-secrets to 0.14.3


macisamuele (approved):
Changes looks good to me.

Windows test failure seems likely to be related to some network issue.
Rerun tests and merge if they are green ;)

Issue #51, opened 2019-07-23
OpenAPI 3.0 compatibility


joaobiriba:
It could be a good thing if the builder could parse OpenAPI 3.0 swagger json file

https://swagger.io/docs/specification/about/

cortinico:
> It could be a good thing if the builder could parse OpenAPI 3.0 swagger json file

Hey @joaobiriba 
Thanks for the request. Ideally we should bump the Swagger Codegen version we are using to `3.x`.
Unfortunately this introduces a lot of breaking changes with the current [KotlinGenerator](https://github.com/Yelp/swagger-gradle-codegen/blob/master/plugin/src/main/java/com/yelp/codegen/KotlinGenerator.kt) class.

So ideally is something that will be implemented but not in the near future.
We're welcoming external contributors anyway :)

joaobiriba:
Ok, let's see if I can found some time to work on it. This helped me to save some time, I could give back :)

hybras:
So at this time the plugin does not support OpenAPI 3? Or is it json specifically?

cortinico:
> So at this time the plugin does not support OpenAPI 3? Or is it json specifically?

Correct, at this time the plugin doesn't support OpenAPI 3. We might achieve this by bumping the `swagger-codegen` dependency to 3.x but we still have to investigate the amount of work needed.

stephen-mojo:
It has been a while since the last comment on this thread. Has any work been done to bump Swagger Codegen to OpenAPI 3.0? Thanks.

cortinico:
> Has any work been done to bump Swagger Codegen to OpenAPI 3.0? Thanks.

Not on my end. I'm unsure if @macisamuele did some work on that front.

macisamuele:
I've looked into it long time ago (sorry for that) and seems that moving to the latest io.swagger:swagger-codegen (which supports openapi v3) might require some effort and it might interfere a bit (due to templating conflicts) with #103 .
I've sync a bit with @cortinico and we thought that I can start doing some prep work on enabling open api v3 and then afterwards dealing with eventual conflicts.

Enabling polymorphism seems being an interesting feature (at least it is for me) but we are not getting many requests on that regard, so it does not seem wise to hold openapi v3 longer

jmgarbowski:
Hi everyone, is there any updates for this request?

cortinico:
[...]

PR #143, opened 2020-10-24
Bump io.swagger:swagger-codegen to 2.4.16

macisamuele:
As the title says.

The goal of this is to prepare a branch to address the addition of OpenAPI v3 (#51) and ideally we might (even if I doubt) reduce some conflicts.

Still bumping a dependency is not a bad idea, especially if tests continue to be green


macisamuele:
Ps.the title is inaccurate...the bump is to patch 16

cortinico (approved):

PR #145, opened 2020-11-18
Initialize optional query parameters with null

agologan:
Pretty self explanatory, helps with calling APIs with many parameters as you don't have to specify all params.

_Considered also initialising with default value but that's more like the value the server assigns when a param is missing so I don't think it's wise to pre-initialize it in the client._


cortinico (approved):
Change looks fine to me 👍 I'd love to get an ack from @macisamuele as well.
Moreover, @agologan can I ask you to add a test here https://github.com/Yelp/swagger-gradle-codegen/blob/master/samples/junit-tests/src/test/java/com/yelp/codegen/generatecodesamples/ValidParameterTest.kt for a call to `getSymbolsInParameterName()` with no parameters?

agologan:
> [...] can I ask you to add a test [...] for a call to `getSymbolsInParameterName()` with no parameters?
done.

cortinico (approved):

macisamuele (approved):
Looks good to me.

Thanks a lot for contributing.

@cortinico will you follow-up with merging (and eventually releasing) this?

cortinico:
> @cortinico will you follow-up with merging (and eventually releasing) this?

Let's merge this.
For the releasing part, I'd love to have this into 2.x
I'll have some time during the XMas eve to review/develop all the remaining features for 2.x

PR #153, opened 2021-08-08
Bumping a lot of dependencies

cortinico:


macisamuele (approved):
lgtm given we have green tests

